Issue #5, opened 2021-07-16
Azure Blob storage SDK dependency injection in `Startup.cs`


LarsKemmann:
### Requirements
1. Provide a `BlobServiceClient` that is configured based on the current environment (dev = Azurite, test, prod).
2. Configuration is via `appsettings.json`/.NET Core configuration stack

PR #20, opened 2021-07-16
Azurite guidance and config for _Startup.cs_

LarsKemmann:


LarsKemmann:
This completes #13 and #5.

LarsKemmann:
@davidmalament FYI, I just added the `BlobServiceClient` to _Startup.cs_ along with guidance in the _README.md_ for using the Azurite storage emulator. I've also turned on CI builds with unit tests for all PRs, thanks again for the reminder! 😊

Issue #6, opened 2021-07-16
Provide ability to run Azurite locally w/ test data auto-regeneration


LarsKemmann:
### Requirements
1. _CareTogether.TestData_ needs to expose a `TestDataCleanup` class that can, given a `BlobServiceClient` reference, clear out the test data organization's container and recreate it.
2. Update the _README.md_ file to explain how to run with Azurite for local testing.
3. Configure the Azurite connection string in _appsettings.Development.json_ for development use.

LarsKemmann:
Once #1 is completed, this will be needed to reduce development friction -- so people don't have to manually clear out their Azurite data folder before each debugging session. Otherwise the `PopulateTestDataAsync` method will add a bunch of duplicate events to each event log at each startup, thereby corrupting the logs (and probably crashing the app 😝).

PR #21, opened 2021-07-16
Added test data reset capability to Startup.cs

LarsKemmann:


LarsKemmann:
This completes #6. @davidmalament FYI, this will support integrating the work you're doing (see description of #6 for more info). Check out the `TODO` comments that I added with this PR for guidance on where you can include the new durable event log implementation in _Startup.cs_ so that it can be used for dependency injection.

PR #28, opened 2021-07-18
Issue 1: draft commit for review

davidmalament:


src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,96 @@
...
+                    while ((line = reader.ReadLine()) != null)
+                    {
+                        lineNumber++;
+
+                        T item = JsonConvert.DeserializeObject<T>(line);
+
+                        yield return (item, lineNumber);
+                    }
+
+                    _tenantEventCounts.AddOrUpdate((organizationId, locationId), 0, (OldKey, OldValue) => OldValue + ((int)lineNumber));
```
LarsKemmann, line 91: You can safely do this after the `foreach` loop, just once, with the final `lineNumber` so you don't have to search & update the dictionary for every event.

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,96 @@
...
+                var appendBlob = tenantContainer.GetAppendBlobClient(blob.Name);
+   
+                var eventStream = await appendBlob.OpenReadAsync();
+
+                string eventString = new StreamReader(eventStream).ReadToEnd();
+
+                using (StringReader reader = new StringReader(eventString))
+                {
+                    string line;
+                    long lineNumber = 0;
```
LarsKemmann, line 80: There's a bug here: the `SequenceNumber` you're returning is not the same thing as a per-log-segment line number. Change this to `eventSequenceNumber` and initialize it outside of the `foreach` loop.

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,96 @@
...
+using Newtonsoft.Json;
+using OneOf;
+using OneOf.Types;
+
+
+namespace CareTogether.Utilities
+{
+    public class AppendBlobMultitenantEventLog<T>
+    {
+        private LogType _logType;
```
LarsKemmann, line 19: Each of these private fields is written to only when the class is initialized in the constructor, so add a `readonly` modifier to them so they can't be changed later (e.g., `private readonly LogType _logType;`).

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,96 @@
...
+        private BlobServiceClient _blobServiceClient;
+        private ConcurrentDictionary<(Guid organizationId, Guid locationId), int> _tenantEventCounts;
```
LarsKemmann, line 21: This will have an initial value of `null`, so you need to initialize it (either here or in the constructor) with a `new ConcurrentDictionary<(Guid organizationId, Guid locationId), int>`.

src/CareTogether.Core/LogType.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+namespace CareTogether
+{
+    public enum LogType
+    {
+        COMMUNITY_EVENT, CONTACT_COMMAND_EXECUTED_EVENT, GOAL_COMMAND_EXECUTED_EVENT, REFERRAL_EVENT
```
LarsKemmann, line 6: Use regular CamelCasing for these names, per C# convention:
```suggestion
        CommunityEvent, ContactCommandExecutedEvent, GoalCommandExecutedEvent, ReferralEvent
```

src/CareTogether.Core/Extensions.cs
```diff
@@ -39,5 +39,18 @@ public static Guid UserId(this ClaimsPrincipal principal)
...
+            { LogType.COMMUNITY_EVENT, "communityEventLog" },
+            { LogType.CONTACT_COMMAND_EXECUTED_EVENT, "contactsEventLog" },
+            { LogType.GOAL_COMMAND_EXECUTED_EVENT, "goalsEventLog" },
+            { LogType.REFERRAL_EVENT, "referralsEventLog" },
+        };
+
+        public static string ConvertToString(this LogType logType)
+        {
+            return logTypeMap[logType];
+        }
```
LarsKemmann, line 54: You can remove these mappings and just use [`Enum.ToString()`](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tostring?view=net-5.0) to generate the log type names.

src/CareTogether.Core/LogType.cs
LarsKemmann: Also, you can change these names to make the generated filenames look friendlier in Blob storage, similar to what you were doing with the dictionary in _Extensions.cs_. Maybe try:
```suggestion
        CommunityEventLog, ContactsEventLog, GoalsEventLog, ReferralsEventLog
```

LarsKemmann (changes requested):
Hey David, thanks so much! This is a great start and it looks like you got many of the requirements implemented. I noticed a couple of issues and some minor things that are more just idiomatic C# usage. Let me know if you have questions about any of these!

src/CareTogether.Core/Extensions.cs
davidmalament: I had originally wanted to do that but for some reason thought the log type string had to be those exact camel case strings I had in the converter. I'm not sure why, but agree this is much cleaner.

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            if(!tenantBlob.Exists())
+            {
+                await tenantBlob.CreateAsync();
+                await tenantBlob.SetHttpHeadersAsync(new BlobHttpHeaders() { ContentType = "application/x-ndjson" });
+            }
+
+            var eventJson = JsonConvert.SerializeObject(domainEvent);
+
+            var cleanEventJson = eventJson.Replace("\n", "").Replace("\r", "");
```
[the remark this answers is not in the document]
davidmalament: Good call. `SerializeObject` has a constructor that takes a `JsonSerializerSettings` which has a `StringEscapeHandling` property. By default this is set to escape newlines, but it also has other options. Do you think the default is good enough, or should we go with one of the others to cover more scenarios? https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_StringEscapeHandling.htm
https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConvert_SerializeObject_3.htm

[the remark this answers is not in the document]
LarsKemmann: The default should be fine, since the only thing that would break the file format is control characters.

src/CareTogether.Core/LogType.cs
davidmalament: My Java background rearing its head here :)

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,130 @@
...
+                        eventSequenceNumber++;
+
+                        T item = JsonConvert.DeserializeObject<T>(line);
+
+                        yield return (item, eventSequenceNumber);
+                    }
+                }
+            }
+
+            _tenantEventCounts.TryAdd((organizationId, locationId), (int)eventSequenceNumber);
```
LarsKemmann, line 113: I think `TryAdd` is incorrect here. It'll only succeed if the key does not already exist ([docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.tryadd?view=net-5.0)). You can just set the entry with an indexer:
```suggestion
            _tenantEventCounts[(organizationId, locationId)] = (int)eventSequenceNumber;
```

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        private async Task<BlobContainerClient> createContainerIfNotExists(Guid organizationId)
+        {
```
LarsKemmann, line 117:
```c#
var orgContainer = _blobServiceClient.GetBlobContainerClient(organizationId.ToString());
await orgContainer.CreateIfNotExistsAsync();
```

```c#
private readonly ConcurrentDictionary<string, bool> _containerExistenceCache = new();
```

src/CareTogether.Core/Utilities/AppendBlobMultitenantEventLog.cs
LarsKemmann: As you pointed out, we can just remove this altogether.

LarsKemmann (approved):
Looks great, thank you @davidmalament!

PR #57, opened 2021-08-21
Add in BottomNavigation to support mobile platforms.

DacoGT:
Use useMediaQuery and breakpoints to determine screen width.


LarsKemmann (approved):
Thanks for cleaning this up! Nice work. :)

PR #103, opened 2021-08-25
Added new Volunteer Component with App Bar to Approvals, Applications…

DacoGT:
…, Progress

This is a draft to show my progress so far. I want to be sure I'm on the right track.


LarsKemmann:
Thanks @DacoGT! I'm working through the review now.

src/caretogether-pwa/src/App.tsx
```diff
@@ -198,40 +129,7 @@ function App() {
...
-              />
-            </div>
-            <IconButton color="inherit">
-              <Badge badgeContent={4} color="secondary">
-                <NotificationsIcon />
-              </Badge>
-            </IconButton>
-          </Toolbar>
-        </AppBar>
+        <Header open={open} handleDrawerOpen={handleDrawerOpen} />
```
LarsKemmann, line 125: The biggest change I would say you need to make is to give each section component (`<Dashboard />`, `<Volunteers />`, `<Referrals />`) its own header. Rather than putting a header here that then has to do a bunch of route matching to figure out where it's being used (which breaks the single-responsibility principle).

Or, a bit more advanced React coding but a cleaner and easier-to-maintain end result: wrap each section component with a `<Header />` that has two slots: one for the header items and one for the main page. Use [this technique](https://reactjs.org/docs/composition-vs-inheritance.html#containment).

LarsKemmann (changes requested):
Finished my initial review, fixed some merge conflicts & minor cleanup items. There are two issues that came up (I added a comment for each), the biggest being the need to make `<Header />` unaware of where it's being used. Otherwise, yes this is definitely the right approach!

LarsKemmann (approved):

PR #107, opened 2021-09-11
Refactored Bottom Nav into a new Footer Component.

DacoGT:
Refactored Bottom Nav into a new Footer Component.
Used Router to set value of Bottom Nav so it's always in sync.


src/caretogether-pwa/src/Components/Footer.tsx
```diff
@@ -0,0 +1,49 @@
...
+    },
+}));
+
+function Footer(props: any) {
+    const { location } = props;
+
+    const classes = useStyles();
+    const [selectedBottomNavAction, setSelectedBottomNavAction] = useState(0);
+
+    const memoizedPathMap = useMemo(() => {
```
LarsKemmann, line 24: Can this just be a simple array, rather than a memoized function result?

LarsKemmann (approved):
Thank you @DacoGT! This is great!

PR #179, opened 2021-10-30
Issue 127 Moved footer out of main div and set main div's height cond…

DacoGT:
Moved footer out of main div and set main div's height conditionally on isMobile


LarsKemmann:
Thanks so much, this is great to see! 😊 Reviewing now.

LarsKemmann (approved):
Looks good and works locally. 🚀

PR #205, opened 2021-11-09
Upgrade React Router to v6

### LINKED ISSUE #202: Upgrade React Router from v5 to v6
https://reactrouter.com/docs/en/v6/upgrading/v5

fhdaax:
If there is anything broken or any mistake, feel free to request a change.

closes #202

### DIFF
--- a/src/caretogether-pwa/src/App.tsx
+++ b/src/caretogether-pwa/src/App.tsx
@@ -6,7 +6,7 @@ import PermPhoneMsgIcon from '@material-ui/icons/PermPhoneMsg';
 import PeopleIcon from '@material-ui/icons/People';
 import ChevronLeftIcon from '@material-ui/icons/ChevronLeft';
 //import DashboardIcon from '@material-ui/icons/Dashboard';
-import { Route, Switch, Redirect, BrowserRouter as Router } from "react-router-dom";
+import { Route, Routes, BrowserRouter as Router, Navigate } from "react-router-dom";
 import { ListItemLink } from './Components/ListItemLink';
 import { Arrangements } from './Components/Referrals/Arrangements';
 import { Referrals } from './Components/Referrals/Referrals';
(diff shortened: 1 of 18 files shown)


src/caretogether-pwa/src/App.tsx
```diff
@@ -130,19 +130,21 @@ function App() {
       <CssBaseline />
       <Router>
         <Header open={open} handleDrawerOpen={handleDrawerOpen} />
-        {isMobile ? null :
+        {isMobile ? null : (
```
DacoGT, line 133: Looks like we might want to get some file format convention going with prettier, a VS Code extension, or perhaps use a githook oncommit

fhdaax: I don't understand what you are proposing. I use prettier, but at the time working in this app, I had disabled it.
I don't see a large change in the code you mentioned. I think the parentheses were recommended.

Do you want me to implement `prettier` that formats code on save, commit, push, or anything like that?

DacoGT: You're right, it's not a large change. File formatting differences can make it difficult to find the intentional code changes. I wanted to bring up the idea to @LarsKemmann. I wouldn't recommend any changes until Lars jumps in.

LarsKemmann: I agree, it would be nice to normally have formatting differences in dedicated PRs for that. Not a problem here and I agree with the changes that were made being recommended. I've created #207 to track implementing Prettier to format code on save.

src/caretogether-pwa/src/Components/Referrals/Referrals.tsx
```diff
@@ -1,23 +1,15 @@
...
-        </Route>
-        <Route path={`${match.path}`}>
-          <PartneringFamilies />
-        </Route>
-        <Route>
-          <Redirect to={`${match.path}`} />
-        </Route>
-      </Switch>
+      <Routes>
+        <Route path='/' element={<PartneringFamilies />} />
```
LarsKemmann, line 9: Nested routes!! 🙌🥳

LarsKemmann (commented):
@fhdaax Thank you so much for putting this together and updating it all the way through the app. The code changes look good! I'm going to pull down the code and make sure it runs, then I'll come back and approve.

LarsKemmann (changes requested):
The overall refactoring looks good, but there seem to be some bugs:
1. Opening the app (`http://localhost:3000`) doesn't handle the base route correctly.
image
2. Browsing to `/volunteers` doesn't work either. Interestingly enough, `/referrals` works just fine.
image

LarsKemmann:
Since the requested changes haven't been addressed in the last two months and the codebase is changing (meaning merge conflicts in this PR will only get worse over time), I've decided to abandon this and replace it with an updated PR.

PR #456, opened 2022-05-15
455 breaking out main drawer to its own component

davidmalament:


LarsKemmann (approved):
This is great, thank you @davidmalament! 👏🏻

PR #457, opened 2022-05-15
438 sorting arrangments by started date desc then create date desc. m…

davidmalament:
…odifying test data to allow testing this issue.


LarsKemmann (approved):
Thank you, especially for the test data! 🎉

PR #469, opened 2022-05-20
373 partnering families expand collapse

davidmalament:
Experimented with a different way of displaying arrangements when collapsed. Let me know if you don't like it and I'll go back to the way it's described in the issue.

### DIFF
--- a/src/caretogether-pwa/src/Components/Referrals/PartneringFamilies.tsx
+++ b/src/caretogether-pwa/src/Components/Referrals/PartneringFamilies.tsx
@@ -1,5 +1,5 @@
 import makeStyles from '@mui/styles/makeStyles';
-import { Fab, FormControlLabel, Grid, Switch, Table, TableBody, TableCell, TableContainer, TableHead, TableRow, useMediaQuery, useTheme } from '@mui/material';
+import { Fab, FormControlLabel, Grid, Switch, Table, TableBody, TableCell, TableContainer, TableHead, TableRow, Tooltip, useMediaQuery, useTheme } from '@mui/material';
 import CircleOutlinedIcon from '@mui/icons-material/CircleOutlined';
 import CheckCircleIcon from '@mui/icons-material/CheckCircle';
 import AccessTimeIcon from '@mui/icons-material/AccessTime';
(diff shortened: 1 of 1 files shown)


LarsKemmann:
This is fantastic! I did a quick test and it looks great. I have a few suggestions, I'll add those as soon as I get back early afternoon.

LarsKemmann:
Sorry, yesterday did not go as planned (car troubles). Here's what I'm thinking.

**Observations:**
- The icon approach is a fantastic way to compress the information.
- The contrast on some of the icons needs to be improved.
- It's hard to tell, when only a few icons are present, what the different possible states are.
- There's no differentiation between different arrangement _types_, which is important (child hosting vs. mentoring, for example).

**Suggestions:**
- Split the single "Arrangements" column into one column per arrangement type defined in the policy.
- When there are no arrangements of a given type at a particular state (e.g., no Friendings in Ready to Start state), show the icon for that state (without a number, i.e. no "0") so that the user is able to identify how many possible states there are and what they might mean.
- Use low-contrast gray when the count is 0.
- Use higher-contrast colors (the green and blue are great, the yellow is hard to see and the gray for Created needs to be a dark gray) when the count is > 0.
- The filled green circle attracts the most attention. The primary attention should be on arrangements that are Started. Consider the following icons:
  - [`PendingOutlined`](https://mui.com/material-ui/material-icons/?query=wait&theme=Outlined&selected=PendingOutlined) for Created.
  - I think [`AccessTimeOutlined`](https://mui.com/material-ui/material-icons/?query=clock&theme=Outlined&selected=AccessTimeOutlined) works great for Ready to Start, keep that.
  - [`PlayCircleFilled`](https://mui.com/material-ui/material-icons/?query=play&selected=PlayCircleFilled) for Started.
  - [`CheckCircleOutlined`](https://mui.com/material-ui/material-icons/?query=checkcircle&theme=Outlined&selected=CheckCircleOutlined) for Ended.

LarsKemmann (changes requested):
See suggestions above.

davidmalament:
Good feedback, thanks. I can't find an enumeration of the universe of arrangement types. Is that available somewhere?

davidmalament:
Is it just Hosting, Babysitting, and Friending?

LarsKemmann:
> Good feedback, thanks. I can't find an enumeration of the universe of arrangement types. Is that available somewhere?

It's dynamic, driven by policy (which can be different for each organization & location). Here's the Recoil selector you can read from to get the set of arrangement types:
https://github.com/CareTogether/CareTogetherCMS/blob/51fae7a1df4da15a0feec53733bc7781314e2c30/src/caretogether-pwa/src/Model/ConfigurationModel.ts#L41

That provides the `EffectiveLocationPolicy` instance for the current organization & location, which has a `ReferralPolicy` property that will get you what you need.
https://github.com/CareTogether/CareTogetherCMS/blob/51fae7a1df4da15a0feec53733bc7781314e2c30/src/caretogether-pwa/src/GeneratedClient.ts#L1061-L1064

https://github.com/CareTogether/CareTogetherCMS/blob/51fae7a1df4da15a0feec53733bc7781314e2c30/src/caretogether-pwa/src/GeneratedClient.ts#L1177-L1180

https://github.com/CareTogether/CareTogetherCMS/blob/51fae7a1df4da15a0feec53733bc7781314e2c30/src/caretogether-pwa/src/GeneratedClient.ts#L1290-L1296

davidmalament:
Sorry this took so long...there were a few things that I needed to do some research on to make sure I did them the "right" way. Let me know what you think!

LarsKemmann (approved):
Beautiful. :) And the time you spent researching was a worthwhile investment!!

PR #471, opened 2022-05-22
435 differentiate host families with the same last names in the assignment selection dropdown

davidmalament:
Took some liberties implementing what I feel like is an improvement to the user experience of finding the right individual or family to assign to an arrangement. Happy to go back to the `<Select>` if you don't like it!


LarsKemmann (approved):
I can't tell you how exciting this is. 😃 _**Please**_ feel free to take liberties with the issue descriptions. 😁

image

PR #542, opened 2022-10-08
Arrangement planning

LarsKemmann:


src/caretogether-pwa/src/GeneratedClient.ts
```diff
@@ -4132,12 +4132,15 @@ export class Arrangement implements IArrangement {
     startedAtUtc?: Date | undefined;
     endedAtUtc?: Date | undefined;
     cancelledAtUtc?: Date | undefined;
+    plannedStartUtc?: Date | undefined;
+    plannedEndUtc?: Date | undefined;
```
donschaefer, line 4136: @LarsKemmann - what is the added value of the `| undefined;` declaration given that you're already using a `?` on the property?

src/caretogether-pwa/src/GeneratedClient.ts
```diff
@@ -7425,6 +7445,11 @@ export abstract class ArrangementsCommand implements IArrangementsCommand {
             result.init(data);
             return result;
         }
+        if (data["discriminator"] === "DeletePlannedChildLocationChange") {
```
donschaefer, line 7448: This line (and similar ones above/below) look like they may benefit from the following changes:
- Switch the `if(data["discriminator" === ...)` lines to leverage a switch statement
- Instead of comparing to typed out strings, compare against a string enum (or another mechanism of your choice that would ensure typed-out strings are only ever set in a single place in the application, so as to reduce the potential for human error at a later point in development & potentially assist with intellisense/typing)

src/caretogether-pwa/src/GeneratedClient.ts
LarsKemmann: I'm not sure. _GeneratedClient.ts_ is auto-generated at compile time whenever the generated Swagger file changes.

src/caretogether-pwa/src/GeneratedClient.ts
LarsKemmann: See previous note on auto-codegen process for _GeneratedClient.ts_; otherwise, I 100% agree with you. 😊

Issue #437, opened 2022-05-03
Hide canceled arrangements from view (with option to show)


LarsKemmann:
Currently, all arrangements - including canceled arrangements - are shown on the partnering family screen. This can create visual clutter when one or more canceled arrangements are mixed in with active arrangements. Move the canceled arrangements out of the primary display logic, and hide them behind an [`<Accordion />`](https://mui.com/material-ui/react-accordion/) (before the "New Arrangement" buttons) that, when collapsed, shows the number of canceled arrangements.

https://github.com/CareTogether/CareTogetherCMS/blob/d1c5fa07f344b0e63b90567a776c9f1f019f1e1b/src/caretogether-pwa/src/Components/Referrals/PartneringFamilyScreen.tsx#L278-L298

donschaefer:
@LarsKemmann  - although you've included the relevant lines of code to investigate (enabling someone to possibly provide a code fix without necessarily reviewing the changes they've made in a browser), it'd be helpful if you could provide a link and/or repro steps for navigating to this UI within the app.

Looking through the main screens I'm seeing, nothing is jumping out at me as to where I might find "arrangements" (or what they are for that matter... though this might make much more sense to someone who is already familiar with the organization)

donschaefer:
Looking into this one

LarsKemmann:
@donschaefer You guessed almost what I had intended with interpretation "A", and thanks for thinking it through more thoroughly than I did. 😊 What I had envisioned was, instead of a new section, to simply create a new block within the Masonry component (before the block of "New Arrangement" buttons) that is a collapsed list of cancelled arrangements. However, what I didn't think through was that expanding that accordion would then not tile nicely with the rest of the masonry blocks.

After thinking about it more, and considering the direction I'd like the overall UX to go in the future, the condensed approach (B) is superior. Let's go with that.

LarsKemmann:
This enhancement is now live. Many thanks to @donschaefer for your help on this!!

simply-emily:
@donschaefer This is so helpful! This will really declutter the view of the families that we're serving, especially the ones that have more complex needs or several children. Thank you! 😊

donschaefer:
No problem, happy to help out! :)

Issue #207, opened 2021-11-14
Implement Prettier to format client code on save


LarsKemmann:
Configure https://prettier.io/ in the codebase, at least for client-side code to start.

Here are the preferred initial settings:

- indent with two spaces for TypeScript
- indent with four spaces for C# (if applying Prettier to server-side code as well which is not required in this issue)
- in general, prefer shorter syntax (e.g., don't specify types when the compiler can infer them)
- in general, **stick with the conventions that are most common in the codebase currently** - the goal is ensuring consistency now and making future changes easier

PR #544, opened 2022-10-21
Issue #437 - Add Ability to Hide Cancelled Arrangements

donschaefer:


src/caretogether-pwa/src/Referrals/PartneringFamilyScreen.tsx
```diff
@@ -80,6 +82,68 @@ export function PartneringFamilyScreen() {
...
+      }
+      return o;
+    }));
+  }
+
+  const meetsArrangementFilterCriteria = (arrangement: Arrangement): boolean => {
+    // TODO: Reduce the number of times this method fires on the initial page load
+    let result = false;
+    const selectedOptions = arrangementFilterOptions.filter(o => o.selected).map(o => o.text);    
+    if (selectedOptions.length) {
```
LarsKemmann, line 125: I think this `if` is unnecessary -- if `.length` is zero, `.forEach(...)` will just exit without doing anything.

src/caretogether-pwa/src/Referrals/PartneringFamilyScreen.tsx
```diff
@@ -80,6 +82,68 @@ export function PartneringFamilyScreen() {
...
+  const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null);
+  const handleFilterClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+  const handleFilterClose = () => {
+    setAnchorEl(null);
+  };
+
+  const open = Boolean(anchorEl);
+  const id = open ? 'Filter' : undefined;
```
LarsKemmann, line 95: I would recommend renaming `open` and `id` to be a bit more clear in the context of this (admittedly) way-too-large function, but I'm half-guessing that you're thinking about extracting the filter to its own component? That would be better (also because then we're starting to slim down the `PartneringFamilyScreen`).

src/caretogether-pwa/src/Referrals/PartneringFamilyScreen.tsx
```diff
@@ -228,16 +292,48 @@ export function PartneringFamilyScreen() {
...
+                        vertical: 'bottom',
+                        horizontal: 'left',
+                      }}
+                    >
+                      <FormControl sx={{ m: 3 }} component="fieldset" variant="standard">
+                        <FormLabel component="legend">Filter Arrangements</FormLabel>
+                        <FormGroup>
+                          {
+                          arrangementFilterOptions.map((o) => {
+                            return (
```
LarsKemmann, line 321: You actually don't need the `{ return ( /*...*/ ); }` -- you can just do something like:
```
<FormGroup>
  {arrangementFilterOptions.map(o => (
    <FormControlLabel /*...*/ />
  ))}
</FormGroup>
```
(and I'm even not sure if the inner parentheses are needed)

LarsKemmann (changes requested):
Nice work @donschaefer! This runs great locally. I just had a couple notes on code style/organization 1) for consistency, because I still haven't implemented automated code formatting via Prettier (#207), and 2) because this component, like several of the others I had created, is already embarrassingly large. 😳

src/caretogether-pwa/src/Referrals/PartneringFamilyScreen.tsx
donschaefer: Fixed

src/caretogether-pwa/src/Referrals/PartneringFamilyScreen.tsx
donschaefer: This is now fixed & also moved to it's own component over @ FilterMenu.tsx

src/caretogether-pwa/src/Referrals/PartneringFamilyScreen.tsx
donschaefer: Thanks for the suggestion - Fixed

src/caretogether-pwa/src/Families/FamilyScreen.tsx
```diff
@@ -106,6 +108,34 @@ export function FamilyScreen() {
...
+  enum ArrangementFilterOptionLabel {
+    Active = "Active",
+    Cancelled = "Cancelled"
+  }
+  const { arrangementFilterOptions, handleFilterArrangements } =
+    useFilterMenu(Object.keys(ArrangementFilterOptionLabel), [ArrangementFilterOptionLabel.Active]);
+  const meetsArrangementFilterCriteria = (arrangement: Arrangement): boolean => {
+    let result = false;
+    const selectedOptions = arrangementFilterOptions.filter(o => o.selected).map(o => o.text);
+    selectedOptions.forEach((option) => {
```
LarsKemmann, line 121: This code could be simplified to just `const selectedOptions = ....filter(...).map(...).any(o => ...)` (so no `forEach` is needed).

LarsKemmann (changes requested):

src/caretogether-pwa/src/Families/FamilyScreen.tsx
LarsKemmann: The reason to do this is that it would also get rid of the `let result = false` (avoiding mutable state) and the need for things like `break` statements (which can be hard to follow).

donschaefer: Great suggestion - implemented! Also tweaked the UI to collapse when there were no matching results since I couldn't think of a valid use case for persisting empty placeholder space. 

To be clear, it already starts out collapsed if there aren't any results, but... prior to my latest update, if you add arrangements to the mix & then adjust the filters to filter them back out, Masonry was retaining whatever the last height was before the results went from a length of > 0 to a length of 0. This update makes it so that when there are 0 results, it ALWAYS collapses.

That said, I'm not _thrilled_ with the inline style tag I used to accomplish that UX improvement, so... if you have suggestions on a better approach there, I'm happy to take a pointer.

LarsKemmann (approved):
Nice work! I'm good with the inline style tag; at some point soon this screen needs a full design overhaul and the masonry component is one thing I'm hoping to get rid of by swapping it for a more table-based layout with side-sheet editing. (Which is partly why I'm thrilled at how reusable this filtering logic is. :))

Issue #564, opened 2022-12-29
Caching for CombinedFamilyInfoFormatter output (w/ invalidation)


LarsKemmann:
The most significant design aspect here is how to flag which families' data to invalidate.

Moving to a unified RecordsManager first (so that there is a single records-management command pipeline to decorate with the cache) may be the best strategy.

PR #586, opened 2023-01-04
564 Caching for CombinedFamilyInfoFormatter output (w/ invalidation)

LarsKemmann:
https://github.com/CareTogether/CareTogetherCMS/issues/564


LarsKemmann:
This currently suffers from a race condition. Let's revisit once more performance analysis has been completed.

LarsKemmann:
This also wasn't caching user-level disclosure correctly. The cache would really need to be either partitioned per user, or used only for the pre-disclosure calculation portion.

More profiling is needed.

Issue #280, opened 2022-01-25
Clicking off a dialog should not count as closing/cancelling it


LarsKemmann:
This change needs to be applied consistently to every dialog in the client-side codebase. Several dialogs use the `<UpdateDialog />` custom component, so that only requires a single change; other dialogs use the MUI `<Dialog />` component directly and will need to be updated individually.

Guidance can be found in [the MUI documentation for `<Dialog />`](https://mui.com/material-ui/api/dialog/). Essentially, the `onClose` callback needs to filter the `reason` and only perform the close action if `reason !== "backdropClick"`.

LarsKemmann:
To do this right, the Material-UI V5 upgrade should happen first.

LarsKemmann:
This is now ready to be worked on.

donschaefer:
Not diving into this just yet (as it seems it'll likely involve a good bit of individual changes & testing), but wanted to share a link that may provide some value to whoever ends up working on it (even if that ends up being me eventually)...

https://smartdevpreneur.com/how-to-disable-mui-dialog-backdrop-click/

LarsKemmann:
#654 addressed most of this. When I tested it, some dialogs still closed in response to clicking their backdrop (adding a new volunteer family, for example), so this issue needs another PR to fully resolve it.

PR #654, opened 2023-06-04
Fix for issue #280

donschaefer:
Updates to resolve Issue #280


donschaefer:
Temporarily blocked by an inability to run/debug the project locally. Will resume work when I have some more time & can troubleshoot the local issues that are preventing me from moving forward.

LarsKemmann:
Thank you @donschaefer!! If you're having trouble, check out the updated readme file at the root of the project.

LarsKemmann:
@donschaefer How close do you think this draft code is to being correct? Is it far enough along that I can take it from here?

donschaefer:
I do not recall tbh - I'd say if you need to move this forward sooner rather than later, you'd probably be best served by starting fresh. All I recall offhand is that I was playing around with a few things that I thought would provide at least a starting point for a fix, but then ran into issues when I tried to run/test locally & never ended up circling back. 

Honestly, I don't think things had even gotten far enough to be considered a "checkpoint" (the point at which I'd normally commit and/or do a draft PR) & probably wouldn't have posted the draft PR at all had it not been for the project instructions encouraging draft PRs when any work has been started.

donschaefer:
Hey @LarsKemmann - I _think_ I've added code that should achieve the desired behavior for all dialogs at this point, but given that I've only manually tested a handful of them, haven't taken the time to do a second pass for potential optimizations/refinements & haven't yet added any unit tests, I don't think this is ready to exit "Draft" status yet.

With that said though, if you don't feel like waiting for those updates, it should at least now be at a point where you can take it across the finish line.

donschaefer:
@LarsKemmann - apologies for the lack of follow up on this. That said, given how busy things have gotten on my end, I wanted to give you a heads up that I'm unlikely to be able to add unit tests for this work any time soon. With that in mind, I'm going to remove "Draft" status & just suggest that you test it thoroughly and find an alternative route for getting unit tests added for this.

LarsKemmann (commented):
The code updates look good! Sorry about the merge conflicts. 😌 I'll try to minimize those going forward. After you fix the latest (and last for now, I promise!) merge conflict, I'll run a test locally and then it should be good to merge!

donschaefer:
Merge conflicts resolved :)

LarsKemmann (approved):
Nice work on this Don!! It's going to be a big help. I confirmed that it's working on most of the dialogs in the Family screen. I tested the 'Create Volunteer Family' dialog and that one didn't work for some reason (clicking on the backdrop closed it). Would you be able to take a systematic look at all the dialogs in the app when you have a chance and test each one, then submit any needed fixes in a follow-up PR?

donschaefer:
@LarsKemmann - Apologies - I thought I had done a systematic review of the dialogs, but I'm guessing there are some syntax variations that I didn't happen to catch with my initial scans. That said, while I'm happy to take a 2nd pass at this, given your preference for transitioning to the use of fly-out components anyway, would it be more efficient to simply prioritize the dialogs that are still closing with backdrop clicks for that transition as we're able to?

donschaefer:
@LarsKemmann - While the prior suggestion may still be a good one if issues persist, I think I found where things went wrong for several of the dialogs & it basically amounts to a bit of dyslexia on my part in the prior PR... I'll post a new PR shortly to fix those instances (which includes the one you found on 'Create Volunteer Family')!

PR #709, opened 2024-03-16
Flip the backdrop click logic for several dialogs

donschaefer:
@LarsKemmann - This PR should correct the faulty ternary evaluations introduced during #654


LarsKemmann (approved):
This fixes all the issues I saw, thanks @donschaefer!

PR #711, opened 2024-03-17
Convert Adult & Child Edit Dialogs to Drawers

donschaefer:
@LarsKemmann - This PR will update the volunteer family screens to leverage the drawer component (instead of the dialog component) for editing both adults and children & automatically implements a vertical scroll if the drawer height exceeds the screen height

### DIFF
--- a/src/caretogether-pwa/src/Families/AdultCard.tsx
+++ b/src/caretogether-pwa/src/Families/AdultCard.tsx
@@ -30,13 +30,12 @@ import { MissingRequirementRow } from "../Requirements/MissingRequirementRow";
 import { ExemptedRequirementRow } from "../Requirements/ExemptedRequirementRow";
 import { CompletedRequirementRow } from "../Requirements/CompletedRequirementRow";
 import { IndividualVolunteerContext } from "../Requirements/RequirementContext";
-import { useDialogHandle } from "../Hooks/useDialogHandle";
-import { EditAdultDialog } from "./EditAdultDialog";
 import { useCollapsed } from "../Hooks/useCollapsed";
 import { useFamilyLookup } from "../Model/DirectoryModel";
 import { useFeatureFlags } from "../Model/ConfigurationModel";
 import { useDrawer } from "../Generic/ShellDrawer";
 import { ManageUserDrawer } from "./ManageUserDrawer";
+import { EditAdultDrawer } from "./EditAdultDrawer";
 
 type AdultCardProps = {
   familyId: string,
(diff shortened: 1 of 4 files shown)


LarsKemmann:
Note to self: I've got to figure out if there's an easy way for me to resolve merge conflicts for a PR branch from your repo. There might not be.

donschaefer:
@LarsKemmann  - merge conflicts are now resolved

> Note to self: I've got to figure out if there's an easy way for me to resolve merge conflicts for a PR branch from your repo. There might not be.

... oddly enough, the only "conflicts" that showed up when I merged main into this branch were the fact that this branch had 2 new files, so I'd agree... definitely _should_ be a way to automatically handle simple conflicts like that, as I feel like that's something that Azure DevOps (where most of my experience is) wouldn't have complained about

LarsKemmann:
This is a great start! The functionality you've added so far looks solid. A couple of things I'm wondering about:

1. Can you update the drawers so that clicking the backdrop doesn't close them?
2. Now that I'm actually seeing these dialogs as drawers, I'm realizing there are aspects I hadn't thought through the first time around, such as:
  1. Saving, cancelling, and closing - in a drawer, it's conventional for the entire drawer's edit commands to be applied in one go when the user hits 'save'. So, instead of having a bunch of little 'edit' buttons that then morph into 'save' and 'cancel' buttons next to each field, there would be a single pair of 'save' and 'cancel' buttons at the bottom of the drawer, and all the individual fields would automatically be editable (assuming the user has permission to edit them). I'm not convinced that this is the right thing to do but I'm also seriously considering it... and considering a similar shift away from per-field 'save' and 'cancel' editing (basically there are lots of little "modes" in the UI today which add lots of extra clicks and mental-model complexity for the user) and towards more of a traditional batch-like/form-based editing feel (even if the individual edits under the hood are still represented by individual commands being sent to the server).
  2. A title bar for the drawer that provides some context for the user.
  3. And a bit longer-term, I'm still thinking about whether the whole progressive-information-disclosure pattern could be made a lot more consistent throughout the entire app, ideally without having overlapping/modal UI like this at all. In other words, getting away from the many different centered dialogs and shifting to drawers feels like just the first step, and now that I can see what the app feels like in this example without dialogs I'm wondering if we should pause on this and first get some more clarity around the overall vision.

donschaefer:
@LarsKemmann - While I should be able to do your first request, I think 2.1 & 2.3 dovetail nicely with some of the edit/layout consistency concerns I raised in my recent review & would agree that it might be good to hit pause on this until you've had more time to weigh options (which I'm happy to assist with during our scheduled calls if you'd like) & determine how you want to move forward. 

After all, it wouldn't make much sense to spend effort making updates to these dialogs/drawers (or any others) if ultimately we just end up pulling the data/functionality back into the pages themselves.

LarsKemmann:
Closing this without merging to clean up the PR queue. We'll continue working on progressive disclosure-oriented redesign of the family screen and other UI throughout CareTogether to improve usability and consistency. This was a valuable experiment; it showed that the existing patterns in the UI (even the more 'polished' ones) are not enough to solve the problems and that a broader redesign is required.

PR #715, opened 2024-03-20
Approval status updates

LarsKemmann:


LarsKemmann:
@donschaefer FYI, you might cringe at the color scheme (I blame defaults 😅) but I just added this capability to make it much easier to tell the different status values apart.

LarsKemmann:
@donschaefer You should be able to reuse some of the same VolunteerRoleApprovalStatusChip-related logic from here when updating the Approvals screen, it's the same logic as on the FamilyScreen and AdultCard.

Issue #714, opened 2024-03-19
Remove/Refactor Columns to Improve Readability


donschaefer:
1. Combine first/last name columns
2. Remove "Gender" & "Age" columns
3. Make family names bold
4. Combine All remaining columns into a single "Roles" column that extends the remaining width of the page. Within that column, "Host Family" role/status should be displayed at the family level & all other roles/statuses should be displayed in a row that wraps to additional lines (thus extending the height of the row itself) when necessary (i.e. when there are many roles present and/or too little room to display them all on a single row due to device width restrictions)
5. Add two filter menus above the volunteers table: Roles and Statuses. The Roles filter gets applied first. Both filters are logical ORs of their options, and both filters have a 'Not Applied' option. The Statuses filter will apply just to the filtered subset of Roles.

donschaefer:
@LarsKemmann - While I do appear to have the ability to create an issue, I lack the ability to...

1. Create Labels / "Anyone with write access to a repository" - https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#creating-a-label)
2. Apply Labels / "Anyone with triage access to a repository can apply and dismiss labels" - https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label

With that said, I'll hold off on creating additional issues for the time being & circle back later when additional permissions are available

LarsKemmann:
> @LarsKemmann - While I do appear to have the ability to create an issue, I lack the ability to...
> 
> 1. Create Labels / "Anyone with write access to a repository" - https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#creating-a-label)
> 2. Apply Labels / "Anyone with triage access to a repository can apply and dismiss labels" - https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label
> 
> With that said, I'll hold off on creating additional issues for the time being & circle back later when additional permissions are available

I think I fixed it, try again?

LarsKemmann:
I just realized that this issue will need to be paired with an update to how we do filtering on this screen for the different roles. For now, maybe the easiest is to just have a two level nested menu for filtering Roles? I want to give that some more thought though.

donschaefer:
[...]

PR #723, opened 2024-03-23
Completion of Request Items 1-3 for Remove/Refactor Columns to Improve Readability #714

donschaefer:
This PR resolves the first 3 action items outlined in #714 (combine the name columns, remove gender/age columns & make family names bold) on both Volunteers screens (approvals & progress)


donschaefer:
Hey @LarsKemmann - Keeping this in draft mode for now... partially because the UI is still nasty & partially because I know I'm close on the functionality, but I have a few questions I'd like to get your feedback on before going any further, namely:

1. Do you want dates displayed on the role chips (as they are elsewhere) or do you want the dates omitted for the volunteers screen?
2. I noticed that, if for example, I filter "Family Friend" down to "Prospective" the current behavior is to show ALL family members that belong to a family that has at least 1 "Prospective" family friend within it (as opposed to ONLY showing the family members who themselves are "Prospective" family friends)... would you like to keep this behavior or would you like to modify it to only show the individuals that match those individual roles? 
3. When a user checks the "Not Applied" checkbox, how would you like that to work... asking because I'm noticing that there are both "-" values for legitimate unknowns & "??" values for unhandled use-cases & I'm wondering if the "Not Applied" should act as a "catch-all" for everything that doesn't match a legit value or if it should only match on truly null/undefined scenarios.
4. Do you have any thoughts/preferences I should use as guidance when determining how to style/position/label the drop-downs for final release?

**Note**: you can pull down the branch & test it if you'd like, just try to ignore the old columns/code that I left in place for reference purposes until I'm ready to finalize things & clean that stuff up

LarsKemmann:
1. Let's keep dates displaying for now.
2. Let's keep showing the whole family for now (but only the matching role chips). This could be something we revisit.
3. Use Not Applied as a catch-all (both null and unknown values). The unknown values should be an exception case so it doesn't need to be handled very cleanly, it just needs to be visible (via the ??) when it happens.
4. Above the table, preferably with the other toolbar controls on the right.

I'll test shortly.

donschaefer:
Hey @LarsKemmann - I put 5hrs into this this evening & while I was able to make some progress, it has been much slower going. I won't go so far as to say that I definitely can't finish it this week, but I'm skeptical I'll be able to finish it within another 3hrs... I added a bunch of TODO comments in the file & some additional notes that may help you troubleshoot if you want to take a look where things are currently at.

LarsKemmann:
Thank you so much for the update @donschaefer -- will look into it this morning! I'm not concerned about the hours on my end, but I totally understand you can only do what you can do.

donschaefer:
@LarsKemmann - After an additional 2.75hrs, this is now in a much better place, with most filtering scenarios now appearing to work as expected. The two I found that still need attention are:
1. "Host Family" Role (or any other valid) + "Not Applied" Status should return 3 matches, not 0
2. "Not Applied" Role + Any Valid Status should return 0 matches, not all

I will try to circle back to these at some point on Saturday to address them, but other than those 2, I think the only thing left would be some cleanup tasks that I'm holding off on for now until I'm finished troubleshooting.

donschaefer:
@LarsKemmann - This should now be ready for review (an additional 1.5hrs was spent on this between the prior push & the most recent one)

LarsKemmann (approved):
Great work @donschaefer -- I tested with local as well as live data and it behaved perfectly! I found one small problem (display of individual roles is missing in the 'collapsed' mode now) that I'm opening a separate issue for.

Issue #730, opened 2024-04-07
Volunteer Approvals: display of individual roles is missing in the 'collapsed' mode


LarsKemmann:
With the combined 'Roles' column, in the collapsed mode the individual role chips aren't showing. TBD exactly how this should be handled but my gut instinct is that the 'Roles' column in the collapsed mode should show _all_ role chips for the family (i.e., family roles and individual roles for each family member), with the person's first name prefixed before any individual role chips.

The rationale is that people frequently will filter the Approvals screen in collapsed mode, and if so the current display can show matching family rows with nothing showing in the Roles column. (E.g., switch to collapsed mode, then filter on 'family coach' and 'family friend', and you'll see the Emily Coachworthy family showing up as a match but the Roles column is empty so you don't know what role Emily -- or maybe someone else in the family -- matched on.)

LarsKemmann:
@donschaefer This is the follow-up issue I mentioned that I would create when I approved & merged #723.

LarsKemmann:
I'm going to mark this as Dev Ready based on assuming that the design approach I suggested above (my "gut instinct" idea) is the correct one. @donschaefer I'm open to any other ways you think this should be addressed, but I do think this takes priority over other issues in the Ideas column.

donschaefer:
@LarsKemmann - while I can certainly make the adjustments you're describing, I have to confess that I didn't notice the functionality was working that way previously & hadn't thought to test for that and/or make it work that way because it never occurred to me that we'd want to put individual roles into the family area (even if it's only when the view is collapsed) as I would personally find that experience confusing... 

I'll grant you it was likely less confusing when the various roles had their own columns & It may very well be that existing users would be expecting that if that's how things were working previously, but I think a good compromise solution might be the following...

When the view is collapsed:
- Add a "Family" label (or something similar) to the left of the family roles
- Create a new row of chips within the family row, but vertically positioned below the "Family Roles"
- Give the new/second row of chips a "Individual" label (or something similar)

A given family row (when the screen is viewed in collapsed mode) would then have a layout that appears roughly as follows (using old-school table/tr/td structure for communication/pseudo code purposes to clearly communicate the concept... actual code would NOT use these):
`
<table>
<tr>
<th>Family</th>
<th>Roles</th>
</tr>
<tr>
<td>Coachworthy Family</td>
<td>
<table>
<tr><td>Family:</td><td><div>Host Family: Approved until 4/1/25</div></td></tr>
[...]

PR #735, opened 2024-04-18
Issue #730 - Display Individual Roles on Collapsed Family Rows

donschaefer:
@LarsKemmann - Here's an initial pass at resolving issue #730 like we had discussed, but it's worth mentioning that I think this is one of those cases where having a bunch of extra test data locally would really help me to be able to fine-tune the UX a bit better before pushing up a PR.

That said, I'm sure it'll be an improvement for your users either way, so feel free to merge if it looks good to you & I can always circle back at a later point to adjust margins, padding, etc. when I've got more data to work with.


LarsKemmann:
Thanks so much @donschaefer! This is definitely good enough to ship now. I'll work to dedicate some time to additional test data in May.

LarsKemmann (approved):

PR #743, opened 2024-05-03
update the logo icon

Rabi94:
The issue was that the icon on the app page in the Edge browser had a white background. To fix it, I found that the problem was with the favicon image—it wasn't truly transparent. I edited the image to make it transparent and then re-uploaded it to the project. Everything seems to work fine on my end. Could you please review it.


LarsKemmann (approved):
Perfect, looks great - thank you! 😊

Issue #741, opened 2024-04-30
Ability to select the referral to view on the family screen


LarsKemmann:
Extracted from #739:
> allow the user to see a list of past referrals and select one referral to view the details of, with the most recent referral being the default.

PR #745, opened 2024-05-28
Feature/741 referral selection

donschaefer:
@LarsKemmann - please review this PR to address issue #741 when you have some time. While testing, please pay special attention to...
1. the "Incomplete" section (as it's hard to tell whether or not it's getting updated appropriately when switching referrals with the test data available) & 
2. the "Documents" section (given that it _appears_ as though documents are meant to persist with the family rather than being referral specific, so I _think_ they're not meant to be updated, but I'm not positive that's an accurate assumption)

### DIFF
--- a/src/caretogether-pwa/src/Families/FamilyScreen.tsx
+++ b/src/caretogether-pwa/src/Families/FamilyScreen.tsx
@@ -112,11 +112,11 @@ export function FamilyScreen() {
   }
 
   let referralRequirementContext: ReferralContext | undefined;
-  if (family?.partneringFamilyInfo?.openReferral) {
+  if (selectedReferral) {
     referralRequirementContext = {
       kind: "Referral",
       partneringFamilyId: familyId,
-      referralId: family.partneringFamilyInfo.openReferral.id!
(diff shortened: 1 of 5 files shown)


LarsKemmann:
@donschaefer I started testing with live data and ran into two issues. One, I haven't been able to isolate yet, but the other I can reproduce with the test data: open the "Emmett Brown" partnering family and choose 'Close Referral'. I get the following error:
image

donschaefer:
> @donschaefer I started testing with live data and ran into two issues. One, I haven't been able to isolate yet, but the other I can reproduce with the test data: open the "Emmett Brown" partnering family and choose 'Close Referral'. I get the following error: image

This particular issue should now be resolved @LarsKemmann. Please let me know what you find in terms of the other intermittent issue you were seeing.

LarsKemmann:
@donschaefer I haven't been able to reproduce the other issue so I'm going to merge this. **Thank you** for all your efforts on this one! It's going to be a big help -- a seemingly small but functionally very important step in the direction we want to go.

LarsKemmann (approved):

PR #751, opened 2024-06-14
Search Box Update

Rabi94:
I updated the search box design to be centered on both desktop and mobile. I also enhanced the search functionality so you can find family members using the adults names, their emails, phone numbers, and addresses. Additionally, it now includes results for children as well.


LarsKemmann:
This is fantastic @Rabi94!! One thing -- can you remove all the commented-out code, and remove "TODO:" from the comments in the `filterFamilies` function? Also there's a "Next steps" comment at the bottom of _ShellAppBar.tsx_, you can remove that as well.

LarsKemmann (approved):
This looks great and works when I tested it -- thank you @Rabi94!

LarsKemmann (approved):
Works even better after fixing an edge case when two or more families have the same name

PR #764, opened 2024-08-05
Add Prettier

PabloDinella:
I added prettier with some basic config for the js/ts files. VS Code will automatically use prettier for formatting, and eslint for linting, regardless of the users default settings.

Added the prettier plugin for eslint to avoid rules conflict, leaving formatting to prettier.

Formatted all files in src in a separate commit to facilitate review (you can review changes by commit): 
image


LarsKemmann (approved):
Thank you! The only issue was the codegen file (_GeneratedClient.ts_), I excluded that to prevent noise in the file history.

PR #766, opened 2024-08-05
Change h3 to typography

Rabi94:
I changed all the h3 elements to Typography components with variant="h3" and set the font size to 2rem to maintain the original font size.

### DIFF
--- a/src/caretogether-pwa/src/theme.tsx
+++ b/src/caretogether-pwa/src/theme.tsx
@@ -1,5 +1,6 @@
 import { createTheme } from '@mui/material/styles';
 import { amber } from '@mui/material/colors';
+import { fontSize, typography } from '@mui/system';
 
 export const theme = createTheme({
   palette: {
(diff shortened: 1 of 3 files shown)


LarsKemmann (changes requested):
The change to `<Typography>` is good, and we want to use the `h3` variant. But, the font size is too big and we don't want to set it everywhere. Please set it in the global Material UI theme settings, and match (roughly) the font size for `h6`.

LarsKemmann:
https://v5.mui.com/material-ui/customization/theming/
https://v5.mui.com/material-ui/customization/typography/

src/caretogether-pwa/src/theme.tsx
```diff
@@ -10,4 +11,14 @@ export const theme = createTheme({
       main: amber//'#ffc400'
     }*/,
   },
+  typography: {
+    h3: {
+      fontSize: '1.3rem',
+      '@media (min-width:600px)': {
+        [createTheme().breakpoints.up('sm')]: {
```
PabloDinella, line 18: I think you forgot to remove the line 17, as both 17 and 18 do the same thing

PabloDinella (changes requested):

LarsKemmann (approved):
Looks great @Rabi94 and @PabloDinella - thank you!

PabloDinella (approved):

LarsKemmann (approved):

LarsKemmann (approved):

Issue #558, opened 2022-12-02
Allow less strict (date-based) monitoring requirement deadlines


LarsKemmann:

LarsKemmann:
Don't remove the time component but only use the date component for overdue calculations

Debi-AVL:
I think this is the same issue I wanted to comment on. I'd love it if the fixed dates on required tasks such as "check in with X family" could be more flexible such as a date range. If a parent ally needs to meet with a parent during the week of July 4-10, it would be helpful to choose the date it happened rather than say it happened on the specific date indicated (such as July 5) and then have to explain in the notes that it was actually on July 6th. I hope that makes sense!

LarsKemmann:
> I think this is the same issue I wanted to comment on. I'd love it if the fixed dates on required tasks such as "check in with X family" could be more flexible such as a date range. If a parent ally needs to meet with a parent during the week of July 4-10, it would be helpful to choose the date it happened rather than say it happened on the specific date indicated (such as July 5) and then have to explain in the notes that it was actually on July 6th. I hope that makes sense!

This is definitely related, but I almost think we might want to create a specific policy that lets this be represented more precisely in the system. Right now, the policy is basically "each check-in is due within 7 days from the last check-in" rather than "check-in is due sometime during this 7-day week and the next check-in is due during the following 7-day week, regardless of when in the first week you actually checked in". Can you confirm if you actually want your policy to be the latter?

Debi-AVL:
Yes - the latter policy works better than a specific day for things like Family Care Supervision, Parent Ally Meetings, Updating Parent Goals, etc. The only one where a hard and fast date would be needed is for the 48 hour visit and host family check-ins when a hosting is taking place.

Debi Forester
Family Care Coordinator | Alongside Families
828-333-9520
***@***.***
alongsidefamilies.org<http://www.alongsidefamilies.org>
________________________________

PR #767, opened 2024-08-12
wip fix for #558

PabloDinella:
wip fix for #558

### DIFF
--- a/src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
+++ b/src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
@@ -9,6 +9,36 @@ using Timelines;
 
 namespace CareTogether.Engines.PolicyEvaluation
 {
+    using System;
+    using System.Collections.Immutable;
+    using System.Text;
+
+    public static class ImmutableListExtensions
+    {
+        public static string ToHumanFriendlyString<T>(this ImmutableList<T> list)
+        {
+            if (list == null || list.Count == 0)
+            {
+                return "The list is empty.";
+            }
+
+            var sb = new StringBuilder();
+            sb.Append("List contents: [");
+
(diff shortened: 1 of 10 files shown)


src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
```diff
@@ -9,6 +9,36 @@
 
 namespace CareTogether.Engines.PolicyEvaluation
 {
+    using System;
+    using System.Collections.Immutable;
+    using System.Text;
+
+    public static class ImmutableListExtensions
```
LarsKemmann, line 16: Move `ImmutableListExtensions` to a new file.

src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
```diff
@@ -9,6 +9,36 @@
...
+    {
+        public static string ToHumanFriendlyString<T>(this ImmutableList<T> list)
+        {
+            if (list == null || list.Count == 0)
```
LarsKemmann, line 20: This method can be made more concise and readable using [the `String.Join` method](https://learn.microsoft.com/en-us/dotnet/api/system.string.join?view=net-8.0#system-string-join(system-string-system-collections-generic-ienumerable((system-string)))) and C# string interpolation:

```c#
public static string ToHumanFriendlyString<T>(this ImmutableList<T> list)
{
    if (list == null || !list.Any())
    {
        return "The list is empty.";
    }

    return $"List contents: [{string.Join(", ", list)}]";
}
```

LarsKemmann (commented):
This all looks directionally correct... and if it works in testing, I'll be thrilled that we don't have to completely rewrite the monitoring time gap logic to make this happen (for now at least). 😊

Good work.

PR #768, opened 2024-08-23
Update test data provider

PabloDinella:
I enriched TestDataProvider with some data, such as some assignments to 1 of the Friending arrangements, to ease the testing during development and have more examples.

I also added a one time recurrence monitoring policy called One Time Checkin, to the Friending policy, but there might be better real life example.

It's WIP still because I need to update the unit tests, might need some help to understand it first.


PabloDinella:
I updated the branch according our last conversations. Created new families and volunteer, so existing authorization tests doesn't break. Also used variable names in a more intuitive way, like `fisherFamilyGuid` instead `guid8`, for the new family.

LarsKemmann (approved):
10/10. Humor quality meets or exceeds expectations. 😁

PR #776, opened 2024-08-30
Family screen v2

PabloDinella:
This PR creates a new Family Screen page, so we can start changing the layout and components.

It adds 2 feature flags:

## FamilyScreenV2

Enables the new version of the Family Screen

## FamilyScreenPageVersionSwitch

Enables the page version switch, that is shown in the top right of the page:
image

---

The user can change the page version, in case there's a bug in the new version or its UX is not good. The switch only changes the flag in the current tab. It's persisted in the session storage, so a page refresh doesn't change the page version he is seeing, but opening a new tab or window will load the flag value from the server.


PabloDinella:
Forgot to mention: the initial idea was to use `useSessionStorage`, but changing the state in a component down the tree don't trigger a rerender at the app routes level. The current version of the hook seems to do so, but apparently we use an old implementation (https://github.com/uidotdev/usehooks/blob/90fbbb4cc085e74e50c36a62a5759a40c62bb98e/index.js#L1177).

So I implemented it using recoil.

LarsKemmann (changes requested):

PR #779, opened 2024-08-30
Family screen V2

Rabi94:
I created the family screen Menu tabs


LarsKemmann:
Looks great overall! I'm not sure about the icons. Can you please experiment with the following

1. Overview --> `Home` (the eye looks like a privacy control)
2. Notes --> what you have might be good (I like the clipboard concept), but also try `Notes` and see which is better (what you have is very similar to the folder you have for "Documents" and that could cause confusion)
3. Documents --> `InsertDriveFile`, `Description`, or `Folder` might be better choices; I'm not sure.
4. Tasks --> leave, it's perfect 😊

LarsKemmann (changes requested):
Looks great overall! I'm not sure about the icons. Can you please experiment with the following

Overview --> Home (the eye looks like a privacy control)
Notes --> what you have might be good (I like the clipboard concept), but also try Notes and see which is better (what you have is very similar to the folder you have for "Documents" and that could cause confusion)
Documents --> InsertDriveFile, Description, or Folder might be better choices; I'm not sure.
Tasks --> leave, it's perfect 😊

LarsKemmann (approved):
Looks great, love the icons! :)

PR #800, opened 2024-11-22
Move timezone conversions

PabloDinella:
Now the `src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs` only works with DateOnly.

That means the PolicyEvaluationEngine will now convert all `DateTimes`s used in calculations to `DateOnly`s.

Note on DateTime to DateOnly conversion: the timezone of the location should be taken into account, so there's no day displacement.


LarsKemmann (approved):

LarsKemmann (approved):

LarsKemmann (approved):

LarsKemmann:
I wonder if this branch just became too large for GitHub to understand that I've actually approved it? 😛

PR #801, opened 2024-11-25
Date only frontend

PabloDinella:
This is the last part of the fix for #558. Only render dates (instead of datetime) on monitoring requirements in arrangements.

Example:

image


LarsKemmann (dismissed):
This looks really good. 😊 In the future we'll have to make sure this works across various date formats globally, but I appreciate the simplifying optimization for now.

PR #803, opened 2024-12-10
Fix a case when multiple location changes doesn't result in the right due dates

PabloDinella:
When multiple location changes happens without a pause in the middle (when the child goes to the parent), the calculation ignores the subsequential locations and only calculates the due dates for the first location. This PR changes that so all location changes are taken into account.

Changes in code style were made by csharpier.

### DIFF
--- a/src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
+++ b/src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
@@ -501,7 +501,7 @@ namespace CareTogether.Engines.PolicyEvaluation
             Guid? filterToFamilyId = null
         )
         {
-            var dateRanges = GenerateDateRanges(childLocations, filterToFamilyId).ToImmutableList();
+            var dateRanges = GenerateDateRanges(childLocations).ToImmutableList();
 
             var filteredDateRanges = (
                 filterToFamilyId != null ? dateRanges.Where(item => item.Tag == filterToFamilyId) : dateRanges
@@ -517,10 +517,7 @@ namespace CareTogether.Engines.PolicyEvaluation
             return DateOnlyTimeline.FromOverlappingDateRanges(filteredDateRanges);
         }
 
-        private static IEnumerable<DateRange<Guid>> GenerateDateRanges(
-            ImmutableList<ChildLocation> childLocations,
-            Guid? filterToFamilyId
-        )
+        private static IEnumerable<DateRange<Guid>> GenerateDateRanges(ImmutableList<ChildLocation> childLocations)
         {
             (DateOnly Date, Guid ChildLocationFamilyId)? previousChildLocation = null;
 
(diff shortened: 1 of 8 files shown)


src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
```diff
@@ -528,29 +519,38 @@ ImmutableList<DateOnly> completionDates
 
         private static IEnumerable<DateRange<Guid>> GenerateDateRanges(ImmutableList<ChildLocation> childLocations)
         {
-            DateOnly? startDate = null;
-            Guid? tag = null;
+            (DateOnly, Guid)? entry = null;
```
PabloDinella, line 522: Is it possible to use an anonymous record here?

LarsKemmann: Try `(DateOnly myDate, Guid myId)?`

LarsKemmann: Rename to `previousNotPausedEntry` for clarity

src/CareTogether.Core/Engines/PolicyEvaluation/ReferralCalculations.cs
```diff
@@ -528,29 +519,38 @@ ImmutableList<DateOnly> completionDates
...
+                    continue;
+                }
+
+                if (entry.HasValue && !childLocation.Paused)
                 {
-                    startDate = childLocation.Date;
-                    tag = childLocation.ChildLocationFamilyId;
+                    yield return new DateRange<Guid>(
+                        entry.Value.Item1,
+                        childLocation.Date.AddDays(-1),
```
LarsKemmann, line 536: Add an inline comment to explain this:
> When a child is cared for, the care duration is from the start date of childcare until the next entry's date of childcare minus one day.

What about a child changing location more than once in the same day, which can happen? This would be a negative date range.

LarsKemmann: Step 1: Add a unit test for the multiple-changes-in-one-day scenario (start w/ parent, go to volunteer, go to another volunteer, go back to the first volunteer, then go to the parent). That will introduce some fun wrinkles. 😅

test/CareTogether.Core.Test/ReferralCalculationTests/CreateChildLocationBasedTimeline.cs
```diff
@@ -16,102 +16,142 @@ public void CreateTimelineWithOneGap()
...
+                        ),
+                        new DateRange(DateOnly.FromDateTime(H.DateTime(1, 20))),
+                    ]
                 )
-            ]));
+            );
+        }
+
+        [TestMethod]
+        public void CreateTimelineFilteredByFamilyIdNoPauses()
```
LarsKemmann, line 140: Need another test to test interaction *with* pauses, as well.

LarsKemmann (changes requested):

test/CareTogether.Core.Test/ReferralCalculationTests/CreateChildLocationBasedTimeline.cs
PabloDinella: Done :)

LarsKemmann (approved):

PR #804, opened 2024-12-16
Format all files

PabloDinella:


.editorconfig
```diff
@@ -0,0 +1,447 @@
...
+dotnet_naming_rule.const_field_members_should_be_pascal_case.symbols = const_field_members
+dotnet_naming_rule.const_field_members_should_be_pascal_case.style = pascal_case
+
+dotnet_naming_rule.local_var_and_method_args.severity = error
+dotnet_naming_rule.local_var_and_method_args.symbols = local_var_and_method_args
+dotnet_naming_rule.local_var_and_method_args.style = camel_case
+
+dotnet_naming_rule.private_member_variables.severity = error
+dotnet_naming_rule.private_member_variables.symbols = private_member_variables
+dotnet_naming_rule.private_member_variables.style = pascal_case_with_underscore_prefix
```
LarsKemmann, line 81: Can you change this to apply `_camelCase` instead of `_PascalCase`? I'm not 100% sure that this is the right syntax to do this but it's a starting point:

```suggestion
dotnet_naming_rule.private_member_variables.style = camel_case_with_underscore_prefix
```

.editorconfig
```diff
@@ -0,0 +1,447 @@
...
+dotnet_naming_style.pascal_case.required_prefix =
+dotnet_naming_style.pascal_case.required_suffix =
+dotnet_naming_style.pascal_case.word_separator =
+dotnet_naming_style.pascal_case.capitalization = pascal_case
+
+dotnet_naming_style.pascal_case_with_underscore_prefix.required_prefix = _
+dotnet_naming_style.pascal_case_with_underscore_prefix.required_suffix =
+dotnet_naming_style.pascal_case_with_underscore_prefix.word_separator =
+dotnet_naming_style.pascal_case_with_underscore_prefix.capitalization = pascal_case
+
```
LarsKemmann, line 123: I see, I think you have to define a custom naming style here: `camel_case_with_underscore_prefix` (basically the same as `pascal_case_with_underscore_prefix` but with the capitalization changed)

.editorconfig
```diff
@@ -0,0 +1,447 @@
...
+dotnet_naming_style.camel_case.required_suffix =
+dotnet_naming_style.camel_case.word_separator =
+dotnet_naming_style.camel_case.capitalization = camel_case
+
+dotnet_style_operator_placement_when_wrapping = beginning_of_line
+tab_width = 4
+indent_size = 4
+end_of_line = crlf
+indent_style = space
+max_line_length = 150
```
LarsKemmann, line 134: Can you reduce the maximum line length? Let's say 100, which is much easier for side-by-side code reviews.

```suggestion
max_line_length = 100
```

src/CareTogether.Api/Controllers/ConfigurationController.cs
```diff
@@ -1,70 +1,81 @@
...
+            EffectiveLocationPolicy? result = await _PoliciesResource.GetCurrentPolicy(organizationId, locationId);
             return Ok(result);
         }
 
         [HttpGet("/api/{organizationId:guid}/{locationId:guid}/[controller]/flags")]
-        public async Task<ActionResult<CurrentFeatureFlags>> GetLocationFlags(Guid organizationId)
+        public async Task<ActionResult<CurrentFeatureFlags>> GetLocationFlags()
         {
-            var result = new CurrentFeatureFlags(
-                InviteUser: await featureManager.IsEnabledAsync(nameof(FeatureFlags.InviteUser)),
```
LarsKemmann, line 65: This is one instance where the automatic naming changes are not great since we lose the clarity of which feature flag corresponds to which property, but arguably we could be returning these in a dictionary anyways. I'm fine with this change for now, but maybe when merging this PR we should create an issue to change `CurrentFeatureFlags` to be a `Dictionary<string, bool>` where the initializer uses `nameof` to set the key? We could even generate this collection dynamically by iterating through all the values on `FeatureFlags`.

LarsKemmann (changes requested):
There are just a few formatting changes to make (in the _.editorconfig_) and one follow-up issue to create, otherwise this looks good. 😊

.editorconfig
PabloDinella: In this case we don't follow IDesign standard?

Just for reference:

IDesign C# Coding Standard 3.1 on section 1.1.4:

> 4) Prefix private member variables with m_or _. Use Pascal casing for the
rest of a member variable name following the m_ :
class SomeClass
{
int m_Number;
}

Not sure if it's possible to enforce the `m` prefix

src/CareTogether.Api/Controllers/ConfigurationController.cs
PabloDinella: I'm not sure why it removed the parameter names, but adding it does not show an error/warning, so I think we are good now :)

.editorconfig
LarsKemmann: Yeah, I think in this case we should start with something more conventional -- `int _number` in this example.

There's actually a larger issue as well: there appear to be a couple of areas where the IDesign _.editorconfig_ itself deviates from the IDesign C# Coding Standard. For example, in suggesting `MyType foo = new(...)` syntax which should _never_ be allowed per the standard:
https://github.com/CareTogether/CareTogetherCMS/pull/804/files#diff-2c0a008aa153fd0d1864fe07006720d7f1f6ab3a3cbbe1fd1644f7de7209c5feR131

I've been thinking that it might be wise to look at using something more common/conventional, i.e. [the .NET Runtime C# coding style](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) which has an _.editorconfig_ ([here's the actual current file](https://github.com/dotnet/runtime/blob/main/.editorconfig)) and which is linked to from https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions. The issue with the `= new(...)` syntax changes in particular is troublesome. Let's plan to review this on a call sometime and figure out the best path forward. I know a lot of work has already gone into this PR, and I don't want it to stay out here for any longer than necessary, but I'm a bit concerned about losing code clarity.

PR #808, opened 2025-02-10
Show custom fields in the Volunteers list

Rabi94:
The custom fields were dynamically added to the Volunteers approval table


PabloDinella:
Validated with a new custom field for a different location :)
image

LarsKemmann (changes requested):
This looks *fantastic*, and while testing it against live data it works beautifully. The only issue is that the text-based value suggestions for the custom fields in the filters need to be sorted alphabetically. The one live organization I tested has 42 values for "Home Church" currently, and they all display in semi-random order. 😅

LarsKemmann (approved):
Nicely done 😊

PR #818, opened 2025-02-13
Custom fields filters

Rabi94:
The custom fields filter is working perfectly. Thank you @PabloDinella

### DIFF
--- a/src/caretogether-pwa/src/Volunteers/VolunteerApproval/CustomFieldsFilter.tsx
+++ b/src/caretogether-pwa/src/Volunteers/VolunteerApproval/CustomFieldsFilter.tsx
@@ -1,0 +1,61 @@
+import {
+  CombinedFamilyInfo,
+  CustomField,
+  CustomFieldType,
+} from '../../GeneratedClient';
+import { CustomFieldsFilterSelect } from './CustomFieldsFilterSelect';
+
+type VolunteerFilterProps = {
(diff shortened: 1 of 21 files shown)


src/caretogether-pwa/src/Volunteers/VolunteerApproval/CustomFieldsFilter.tsx
```diff
@@ -0,0 +1,61 @@
...
+      new Set(
+        volunteerFamilies
+          .map((family) => {
+            const field = family.family?.completedCustomFields?.find(
+              (customField) => customField.customFieldName === fieldName
+            );
+            return field?.value;
+          })
+          .filter((value) => value !== null && value !== undefined)
+      )
```
LarsKemmann, line 34: The text-based value suggestions for the custom fields in the filters need to be sorted alphabetically. The one live organization I tested has 42 values for "Home Church" currently, and they all display in semi-random order. 😅

LarsKemmann (changes requested):
This looks fantastic, and while testing it against live data it works beautifully. The only issue I found was the need to sort the custom field filter values alphabetically. I would prefer to keep "Yes","No" in that order (not "No","Yes"), and ideally only the text-based fields would have the available filter values sorted alphabetically.

LarsKemmann (dismissed):
So, so, so good :)

LarsKemmann (approved):

PR #829, opened 2025-02-21
Exemption fix

PabloDinella:
This fixes a bug where a missing requirement was still showing, even if it was exempted.

This scenario explains the problem:
  - Missing requirement completion with date 1/1/25 UTC. We show this date on UI as is, without converting to local time, which in probably all cases (because customers are in US) would show 12/31/24.
  - Marking it as exempted sends the exemption date as 1/1/25 12AM to server, when it should send 1/1/25 5AM (in case customer's time is EST)
  - When the server calculates missing dates, it converts the UTC dates to customers local time (based on `location.timeZone`, which defaults to EST)
  - So the exempted date ends up being 12/31/24, which don't match the missing date.

So the solution was to convert the date to local time before sending to server.

The completion works fine because the date is set from a datepicker, so it's already the local date. In the exemption, we were just getting the UTC date that was sent from the server before.

I will review the other functions as `exemptVolunteerFamilyAssignmentRequirement` to implement the same fix.

Before:

image

After:

image

## Other changes

Also added the exempted date to the UI, and standardized date format to dd/mm/yy.

## Follow up

What should we do with existing exempted dates?

### DIFF
--- a/src/caretogether-pwa/src/Utilities/dateUtils.tsx
+++ b/src/caretogether-pwa/src/Utilities/dateUtils.tsx
@@ -5,3 +5,14 @@ export function formatUtcDateOnly(input: Date) {
     timeZone: 'UTC',
   }).format(input);
 }
+
+export function convertUtcDateToLocalDate(input: Date) {
+  return new Date(
+    input.getUTCFullYear(),
+    input.getUTCMonth(),
+    input.getUTCDate(),
+    input.getUTCHours(),
+    input.getUTCMinutes(),
+    input.getUTCSeconds()
+  );
+}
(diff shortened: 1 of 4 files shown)


src/caretogether-pwa/src/Utilities/dateUtils.tsx
LarsKemmann, line 1: Just a note that I need to run this locally to generate before & after screenshots and verify it works with some live data.

LarsKemmann (approved):

LarsKemmann (changes requested):
After Lars tests, Pablo will expand this to family function assignment-specific requirements (e.g., 'Home Visit' for host families when the child location is changed).

LarsKemmann:
Before fix:
image

After fix, the same data shows as:
image

After fix, unexempting and reexempting fixes the issue:
image

src/caretogether-pwa/src/Utilities/dateUtils.tsx
LarsKemmann: Test looks good! :)

PabloDinella:
> After Lars tests, Pablo will expand this to family function assignment-specific requirements (e.g., 'Home Visit' for host families when the child location is changed).

Done!

LarsKemmann (approved):

LarsKemmann:
Before fix:
image

After fix, the same data shows as:
image

After fix, unexempting and reexempting fixes the issue:
image

PR #833, opened 2025-03-07
Add assignments to volunteer family info and a way to update related aggregates on client

PabloDinella:
Now, when an arrangement is updated with a volunteer assignment, the volunteer family is also returned as a response to atomic record command, this way the client is always updated.

Also included an assignments field in the volunteer family record.


PabloDinella:
Those changes doesn't apply to `ExecuteCompositeRecordsCommand`, do you think I should also updated that?

LarsKemmann (approved):

Issue #813, opened 2025-02-11
Alert when child location is not 'with parent' for closed arrangements


LarsKemmann:
Added for @AGNMM18

Original request: "Show in inbox when child is still listed with volunteer, alert user when closing without returning child to parent"

Show an inbox alert for each closed arrangement that involved childcare, where the child's current (most recent) location is *not* "With Parent". (Note, in the future, we will have to add something like a "Not Reunified" option for tracking purposes. For now, "With Parent" is the only option we will allow.)

![Image](https://github.com/user-attachments/assets/c9f40dfa-e8f2-4d2c-95be-54c4daaf213c)

Secondly, when closing an arrangement, show a browser alert dialog warning the user that the child is not listed as "With Parent". The user can continue through the dialog, but then the inbox alert must show up. The alert text should read:

> WARNING! You are closing an arrangement without returning the child to the parent. Remember to update the child location to 'With Parent' when a child has been reunified with their parent at the end of an arrangement.

PR #835, opened 2025-03-18
Child location alert 

Rabi94:
Now, we have an alert when a child's location is not 'with parent' for closed arrangements. #813


src/caretogether-pwa/src/Model/QueueModel.ts
```diff
@@ -52,11 +60,64 @@ const missingPrimaryContactsQuery = selector<MissingPrimaryContact[]>({
...
+          arrangement.childLocationHistory &&
+          arrangement.childLocationHistory.length > 0
+            ? arrangement.childLocationHistory[
+                arrangement.childLocationHistory.length - 1
+              ]
+            : null;
+
+        return mostRecentLocation
+          ? mostRecentLocation.plan !== ChildLocationPlan.WithParent
+          : false;
```
PabloDinella, line 98: This part can be simplified like this:

```suggestion
        const mostRecentLocation =
          arrangement?.childLocationHistory?.[
            arrangement.childLocationHistory.length - 1
          ];

        return mostRecentLocation?.plan !== ChildLocationPlan.WithParent;
```

PabloDinella:
LGTM, I also did a slight change to avoid looking up the family unnecessarily: https://github.com/CareTogether/CareTogetherCMS/pull/835/commits/5f20ab074d7b45cb077716d8ebebb3a52dcceca4#diff-1e99f8d6b4a44952a89773843210ea570a1df985de50378d0d3d753ce764092fL96-L104

LarsKemmann (approved):

LarsKemmann:
This will be very helpful for ministry staff who are managing hostings, thank you @Rabi94!

Issue #810, opened 2025-02-11
Add assignment links (as arrangement cards) to (volunteer) family screen


LarsKemmann:
Adding for @AGNMM18

Original request: "Searching for hosting with HF as only info"

For families who have the 'is volunteer family' flag set on their record (which should be all families in the 'Volunteers' list, including staff), add an "Assignments" section to their family screen. This should show below the "Communities" section. This section does not have a separate visibility permission check.

In the "Assignments" section, show an arrangement card (in summary view) for each arrangement that the family or any adults in the family have been involved in (past or present doesn't matter). Sort the arrangements in descending order by date requested, with the most recently requested arrangements shown first.

Arrangement cards should click through to the partnering family. 😊

![Image](https://github.com/user-attachments/assets/d5c051a7-bc6a-4d53-b3b6-37f98463550c)

PR #840, opened 2025-03-23
Add assignment links

Rabi94:
Assignment cards had been added to the (volunteer) family screen. #810


src/caretogether-pwa/src/Families/AssignmentsSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+      results.push({ referralId: x.id!, arrangement: y })
+    )
+  );
+  partneringFamilyInfo.openReferral?.arrangements?.forEach((x) =>
+    results.push({
+      referralId: partneringFamilyInfo.openReferral!.id!,
+      arrangement: x,
+    })
+  );
+  return results;
```
PabloDinella, line 61: This code works totally fine, but a good practice is to strive for immutability, which in general avoids many kinds of bugs, so another approach that is worth trying here is to not create a `results` constant beforehand and push, but to create two constants like `closedReferralsArrangements` and `openReferralArrangements`, provide its values with .map (which is immutable) and concat the values before returning.

src/caretogether-pwa/src/Families/AssignmentsSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function AssignmentCard({ assignment }: AssignmentCardProps) {
```
PabloDinella, line 64: This whole component can go in its own file, so there's no ambiguity about what's inside AssignmentsSection.tsx

PabloDinella (changes requested):

src/caretogether-pwa/src/Families/AssignmentsSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const personAndFamilyLookup = usePersonAndFamilyLookup();
+  const familyLookup = useFamilyLookup();
+  const partneringFamilies = useLoadable(partneringFamiliesData);
+  const navigate = useNavigate();
```
LarsKemmann, line 68: Use `useAppNavigate()` instead which will give you the correct URL structure, including the organization and location IDs.

src/caretogether-pwa/src/Families/AssignmentCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+    ? assignment.childLocationHistory[
+        assignment.childLocationHistory.length - 1
+      ].childLocationFamilyId
+    : null;
+
+  const latestLocationFamily = latestLocationId
+    ? familyLookup(latestLocationId)
+    : null;
+
+  const nextPlannedLocationId = assignment.childLocationPlan?.length
```
LarsKemmann, line 62: Remove `nextPlannedLocationId` since the `nextPlannedLocation` (below) already does the same search. Anywhere that you were using `nextPlannedLocationId`, just use `nextPlannedLocation?.childLocationFamilyId`

LarsKemmann (changes requested):

PabloDinella (approved):

LarsKemmann (approved):
Excellent!! Thank you @Rabi94 😊

Issue #854, opened 2025-04-04
Sort the assignments card


Rabi94:
Sort the assignments from newest to oldest, and add a Show More / Show Less button.

PR #856, opened 2025-04-04
sort assignments

Rabi94:
I sorted the assignment cards from newest to oldest and added a "Show More / Show Less" button. #854

### DIFF
--- a/src/caretogether-pwa/src/Families/AssignmentsSection.tsx
+++ b/src/caretogether-pwa/src/Families/AssignmentsSection.tsx
@@ -1,15 +1,25 @@
-import { Grid, Typography } from '@mui/material';
+import { Grid, Typography, Button } from '@mui/material';
 import { CombinedFamilyInfo } from '../GeneratedClient';
 import { AssignmentCard } from '../Families/AssignmentCard';
+import { useState } from 'react';
 
 interface AssignmentsSectionProps {
   family: CombinedFamilyInfo;
 }
 
 export function AssignmentsSection({ family }: AssignmentsSectionProps) {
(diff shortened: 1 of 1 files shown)


src/caretogether-pwa/src/Families/AssignmentsSection.tsx
```diff
@@ -1,28 +1,48 @@
...
+    (a, b) => {
+      const dateA = a.startedAtUtc ? new Date(a.startedAtUtc).getTime() : 0;
+      const dateB = b.startedAtUtc ? new Date(b.startedAtUtc).getTime() : 0;
+      return dateB - dateA;
+    }
+  );
 
   if (assignments.length === 0) return null;
 
+  const visibleAssignments = showAll ? assignments : assignments.slice(0, 9);
```
PabloDinella, line 22: A good idea is instead of "hardcoding" the number 9 here and in line 37, put it in a `const MAX_INITIAL_ASSIGNMENT_ITEMS = 9` in the beginning of the file, this way if we ever need to change this, we change only in one place :)

Using upper case is a common pratctice for constants like this

PabloDinella (approved):
Nice job! There are two little improvements we can make.

Also, it seems the frontend state is not "auto" updating when I create new arrangements, needing a page refresh to get the most updated data. This is part of #833, I will take a look on what's missing there.

PabloDinella (approved):

LarsKemmann:
@Rabi94 FYI, whenever you address a PR review comment with a code update, you should mark that comment as resolved. Otherwise it looks like the comments are still unaddressed when I go to review.

LarsKemmann (approved):
Works great, nicely done!

Issue #853, opened 2025-04-03
Add a print view for a family's notes


LarsKemmann:
For auditing purposes, there needs to be a way to easily view all notes in a family's records in a way that can be printed (e.g., to a PDF).

Ideally, this view should also include rendering the associated completed requirement name, and the associated arrangement type & person being served (if the note is for an arrangement-related completed requirement or child location change), as well as the note metadata (note date/time & author, note approver date/time & approver).

LarsKemmann:
@PabloDinella @Rabi94 A customer is asking for this to prepare for an audit later this month. Can you two work together on this?

LarsKemmann:
https://www.npmjs.com/package/react-to-print might be helpful in building this.

PR #857, opened 2025-04-08
Print notes

PabloDinella:
Tackles #853 

I added the **Print notes** button to the three dots menu:

image

We talked about moving buttons in the topbar to the right side of its respective section titles, but after doing it I find that the UX can get a bit confusing, so I decided to talk more about that later and just put the print notes button in the menu.

When clicking on the button, the browser print dialog opens as this:

image

### DIFF
--- a/src/caretogether-pwa/package-lock.json
+++ b/src/caretogether-pwa/package-lock.json
@@ -34,6 +34,7 @@
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
         "react-router-dom": "^6.24.1",
+        "react-to-print": "^3.0.5",
         "recoil": "^0.7.7",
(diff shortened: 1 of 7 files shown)


LarsKemmann (changes requested):
This seems great but I see a bug in testing. Here's the Coachworthy Family before:
image

And here's after:
image

The same thing happens with client families as well. It looks like the actual notes aren't rendering anymore? Except in print view which actually works great. :)
image

PabloDinella:
@LarsKemmann fixed, it was because the noteLookup is a stateful/mutable function, and as I'm rendering the timeline twice now (one is for actual web view and other for print view) it was messing things up.

LarsKemmann (approved):

Issue #851, opened 2025-04-02
Enable a PostHog feedback survey


LarsKemmann:
https://posthog.com/docs/surveys

Add a 'Feedback' button to the bottom left of the menu (above the copyright) that has maybe a sad/happy emoji and opens the PostHog feedback survey that we create.

The PostHog survey should have the following elements:
1. A message clarifying: "Have feedback about the CareTogether app? We'd love to hear about it! This is **NOT** intended for feedback about your work with [OrganizationName]; please don't share any sensitive information here as this form goes only to the CareTogether developers."
2. Emoji scale of "How are you feeling about CareTogether today?"
3. Free-text field of "What's going well? What can we improve?"

LarsKemmann:
@PabloDinella I think this would be super helpful. Can you review before @Rabi94 starts on it?

PR #858, opened 2025-04-09
The survey button

Rabi94:
I added a survey button to the sidebar, just above the copyrights, and it's working perfectly! 😊 
#851

### DIFF
--- a/src/caretogether-pwa/src/Shell/Feedback.tsx
+++ b/src/caretogether-pwa/src/Shell/Feedback.tsx
@@ -1,32 +1,27 @@
 import SentimentSatisfiedAltIcon from '@mui/icons-material/SentimentSatisfiedAlt';
-import { useState } from 'react';
+import Button from '@mui/material/Button';
 
 export default function Feedback() {
-  const [isHovered, setIsHovered] = useState(false);
-
-  const buttonStyle = {
-    display: 'flex',
-    alignItems: 'center',
-    gap: '8px',
-    padding: '8px 16px',
-    borderRadius: '16px',
-    backgroundColor: isHovered ? '#e5e7eb' : '#f3f4f6',
-    border: 'none',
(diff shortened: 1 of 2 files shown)


src/caretogether-pwa/src/Shell/Feedback.tsx
```diff
@@ -0,0 +1,32 @@
+import SentimentSatisfiedAltIcon from '@mui/icons-material/SentimentSatisfiedAlt';
+import { useState } from 'react';
+
+export default function Feedback() {
+  const [isHovered, setIsHovered] = useState(false);
```
LarsKemmann, line 5: This works but is not the recommended approach for MUI. It requires an extra `useState` which adds complexity and mandates the use of JavaScript to manage something that is really just visual logic -- which CSS is perfectly capable of handling. CSS has the concept of _pseudo-selectors_ for things like this, in particular `:hover`. Here's [the official list of CSS selectors](https://www.w3.org/TR/css/#selectors) (browser support may vary somewhat but most of these are well-supported). Not only that, but setting styles via the `style` prop gives up some of the capabilities of MUI's theming system, like the use of theme tokens. A better approach is to use the `sx` prop which is available on MUI's `<Button>` component, and on most other MUI components as well. Here's an example in the documentation:
https://mui.com/system/getting-started/usage/#superset-of-css

LarsKemmann (changes requested):

LarsKemmann:
A few slight tweaks are needed to the survey wording:

image

1. "This is NOT intended for feedback about your work" wouldn't apply to volunteers or clients. A more general way to say it is:
  > This is ONLY intended for feedback about the CareTogether web app. Please don't share any sensitive information here as this form goes only to the CareTogether developers.
2. Instead of "Let's do this", say "Provide app feedback"
3. `Start Typing ...` should be `Start typing...`

LarsKemmann (approved):

Issue #859, opened 2025-04-11
Sync family query args with selected referral


LarsKemmann:
As a follow-up to #855 we need to update the URL query parameters now whenever a user changes the selected referral (set the new referral ID parameter and clear out the arrangement ID parameter); I'll create a separate issue for that.

If we don't do this and a user navigates to a specific arrangement, then goes to another screen or family, then hits the 'back' button in the browser, they'll be automatically pushed back to the closed referral, which is likely confusing.

A benefit of doing this is that links to families can now be true deep-links, even when a user browses to a specific referral interactively and then shares their current URL.

PabloDinella:
The main goal is clear, but I think there might be a bit of confusion regarding this part:

> If we don't do this and a user navigates to a specific arrangement, then goes to another screen or family, then hits the 'back' button in the browser, they'll be automatically pushed back to the closed referral, which is likely confusing.

Currently it's not happening like this. This is what happens:

1. User clicks the an alert (navigates to closed referral and scroll to arrangement)
2. Use the searchbar to navigate to another family
3. Click the browser back button
4. Navigates back to previous family, but this time the closed referral was not automatically selected, and didn't scroll to the arrangement in the URL. It stays at the open referral.

I see this as a bug, since the referralId and arrangementId are in the URL, it should show the corresponding referral and scroll to arrangement upon clicking Back. If that's confusing from user's POV, a solution is to clear the querystrings upon arriving in the page, this way the back button doesn't select a specific referral. What do you think?

Totally agree on the syncing the querystring with the radio buttons and the deep-linking benefits though :)

LarsKemmann:
Whatever we do, we should not both clear the query string _and_ sync the URL. My preference is to keep the query string, and keep it in sync, to better align with how we will do deep-linking in the future once the family screen has tabs (V2 UI) and drill-down panels.

LarsKemmann:
This bug that @PabloDinella noted (step 4 in [your comment](https://github.com/CareTogether/CareTogetherCMS/issues/859#issuecomment-2797171072)) only happens when using the search bar, not when using the left nav menu or any other method to open a new family.

LarsKemmann:
After review with @PabloDinella let's do this:

1. Sync the `referralId` with the user's selection of the referral
2. Clear the `arrangementId` when the family screen loads (so only a direct click on a deeplink will cause a scroll, not every time that the user goes back to that family URL in their browser history afterwards)

[...]

PR #861, opened 2025-04-11
Sync family query args with selected referral

Rabi94:
Synced family query args with the selected referral, Now, when a referral is selected , the family query arguments reflect that selection correctly. #859

### DIFF
--- a/src/caretogether-pwa/src/Families/FamilyScreen.tsx
+++ b/src/caretogether-pwa/src/Families/FamilyScreen.tsx
@@ -118,6 +118,8 @@ export function FamilyScreen() {
     c.community?.memberFamilies?.includes(familyId)
   );
 
+  const appNavigate = useAppNavigate();
+
   const familyLookup = useFamilyLookup();
   const family = familyLookup(familyId)!;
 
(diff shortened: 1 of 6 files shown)


src/caretogether-pwa/src/Families/FamilyScreen.tsx
LarsKemmann, line 1: Whenever I open a family _without_ a referral parameter, the `FamilyScreen` should update the query arg when it auto-selects a referral (if there is a referral). For example, when I click on any family from the 'Referrals' screen.

LarsKemmann (changes requested):

src/caretogether-pwa/src/Families/FamilyScreen.tsx
Rabi94: Are you saying that when I open a family, it should automatically select the open referral and show it in the link? Or, if the user navigates to the family screen from the referrals screen, the link should include the ID of the referral they clicked on? @LarsKemmann

PabloDinella: Today, the first behavior you mentioned is already implemented. So the only thing missing now is to add the referralId to the querystring, when the open referral is auto-selected, as I understand.

LarsKemmann: Both of you are close but not 100% there. Also, @Rabi94's idea for navigating from the referrals screen actually makes this even better and I think we can build on it. 😊

1. The behavior today when clicking on a family on the referrals screen is to link just to the family. That's fine (although see point #2 below). What I want to happen anytime the user opens a family that _has_ referrals, since we already have logic to automatically _select_ a referral, is to also update the URL query string with that selected referral ID at the same time. So, clicking on a family without specifying a referral in the URL will lead to the URL being updated with the auto-selected referral ID. This is what @PabloDinella is talking about.

2. Whenever the user clicks on an actual arrangement card in the referrals list screen, it would be great to actually deep-link into that referral as well, not just the family.

PabloDinella: That will be a really nice UX improvement :D

LarsKemmann (approved):

PR #863, opened 2025-04-15
Hide Grant Exemption Tab

Rabi94:
Done! After testing, the "Grant Exemption" tab is now hidden for Emily Coachworthy as expected. Here's a screenshot

Screenshot (171)

 everything looks like it's working perfectly 😊


LarsKemmann (approved):
Works well :)

PR #865, opened 2025-04-16
 Arrangements filter

Rabi94:
The arrangement filter now defaults to no selection, so it shows all arrangements by default. You can filter by Active, Canceled, or Ended. I tested the filter, and everything is working perfectly.

 Here are some screenshots
 
 
Screenshot (178)

Screenshot (176)

Screenshot (175)

Screenshot (174)


LarsKemmann:
First, let's make sure that the default is for just 'Active' to be selected. By default, we want to hide cancelled and ended arrangements.

LarsKemmann:
The next thing is to make it clear which filters are applied. Similar to the filter '1 of 3' display on the Volunteers screen header.

PabloDinella:
Hey @Rabi94, I pushed a commit fixing that bug we talked about earlier. I basically moved state control to a single place, instead of controlling the filters state inside the filters component. That way we got rid of the useEffect, that was causing the bug.

I also moved the whole arrangements section to its own component, to unclutter the FamilyScreen component a bit, and created some other files for a custom hook and the function the actually filters the arrangements.

Now, one thing I realized, we probably should use the same component that is used in the Volunteers screen for filtering (`src/caretogether-pwa/src/Volunteers/VolunteerApproval/VolunteerFilter.tsx`), so we keep the UI consistent throught the app, and don't have multiple implementations for the same features (filtering arrangements and filtering volunteers). Unless we want the arrangements filter to be different. Thoughts here @LarsKemmann?

So, for that, we should move the VolunteerFilter to a more generic place and name, inside as `src/Generic/Filters.tsx` for example.

@LarsKemmann I also suggested Rabi to put an empty state text for when there's no arrangements to show for given filters, like this:
image
What do you think?

If that's okay, the only thing missing is changing the wording a little bit to make it more generic, as right now it only works for when there is no active arrangements.

So here as little todo list so we keep track of progress in this PR:

- [ ] Reuse VolunteerFilter.tsx
- [ ] Move VolunteerFilter.tsx to src/Generic/Filters.tsx
- [ ] Update empty state message for arrangements to be more generic (e.g., "No arrangements match your filters").

LarsKemmann:
Try toggle buttons w/ multi-selection: https://v5.mui.com/material-ui/react-toggle-button/#multiple-selection --- you could even show a badge within each button of the # of arrangements in that status (active, ended, cancelled).

Default should be to show all 3 status options (show all 3 toggle options as 'on' by default).

src/caretogether-pwa/src/Referrals/Arrangements/ArrangementsSection/ArrangementsSection.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <Typography
+              className="ph-unmask"
+              variant="h3"
+              sx={{ m: 0, display: 'flex', alignItems: 'center' }}
+            >
+              Arrangements
+            </Typography>
+
+            <ToggleButtonGroup
+              value={selectedFilters}
```
PabloDinella, line 92: The only change I would make is add `  color="primary"`, I think it make the togglebutton visual more consistent with the app
```suggestion
              color="primary"
              value={selectedFilters}
```

PR #866, opened 2025-04-17
 Schedule call support

Rabi94:
I added a support button at the top, and it looks like what you see in the screenshot. The issue I'm facing is that the 'Schedule Call' button appears a few milliseconds before the ticket is rendered. I'm wondering if I should use a useEffect to delay rendering the 'Schedule Call' button until the ticket is fully loaded.

Screenshot (180)

@LarsKemmann 
Also, about the text above the 'Schedule Call' button , just let me know what you'd like it to say, and I’ll update it for you!

### DIFF
--- a/src/caretogether-pwa/src/OpenTicketForm.tsx
+++ b/src/caretogether-pwa/src/OpenTicketForm.tsx
@@ -1,6 +1,15 @@
+import { Typography } from '@mui/material';
+
 export function OpenTicketForm() {
   return (
     <>
+      <Typography variant="h6" gutterBottom>
+        Submit a ticket (faster turnaround)
+      </Typography>
+      <Typography variant="body1" paragraph>
+        Need help with something specific? Sending us a ticket is usually the
+        fastest way to get your issue addressed.
+      </Typography>
       <script
         type="text/javascript"
         src="https://s3.amazonaws.com/assets.freshdesk.com/widget/freshwidget.js"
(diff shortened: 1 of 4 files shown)


LarsKemmann (changes requested):
@Rabi94 Great work!! The wording is perfect. The two changes I'd like to see are:
1. Add some wording above the ticket form as well.
2. Put the options side-by-side in larger (maybe >= `lg`) sizes: ticket form on the left, office hours on the right. We want to encourage people to use the tickets by default. This also makes the ticket form a bit narrower; right now it's unnecessarily wide on larger sizes. It should be maybe 2/3 of the screen (8/12 of the grid).

LarsKemmann:
Just for fun, I experimented with loading `https://support.caretogether.io` in an `<iframe />` rather than linking to it with a button. It might be something to try in the future, but for now it's best to keep the button; the embedded Bookings page just doesn't look very good here. We can revisit that when we add the actual support portal integration with Freshdesk on this page.

Rabi94:
I’ve made the changes you requested @LarsKemmann, Here are the screenshots for both large and small screens, let me know if there’s anything you’d like me to tweak.


Screenshot (182)

Screenshot (183)

LarsKemmann (approved):
@Rabi94 I cleaned up some of the code and the text; since it was mostly just text changes, it was more efficient to just make the changes than to write them up for you to make. I did fix a filename typo as well. Take a look so you know what I changed.

PR #880, opened 2025-05-02
Add LocationUsers to OData controller

PabloDinella:
I have some questions here:

1) From what I understood, we have to set a primary key, is that right?

2) If so, are user ids the same across locations? If so, I guess we must group the user ids like this:

```json
{
  "@odata.context": "http://localhost:5000/api/odata/live/$metadata#LocationUsers",
  "value": [
    {
      "LocationId": "22222222-2222-2222-2222-222222222222",
      "UserIds": [
        "00000000-0000-0000-0000-000000000000"
      ]
    }
  ]
}
```

3) This user id currently is the person id, how do I get the actual user id?

### DIFF
--- a/src/CareTogether.Api/OData/LiveODataModelController.cs
+++ b/src/CareTogether.Api/OData/LiveODataModelController.cs
@@ -26,7 +26,10 @@ namespace CareTogether.Api.OData
 {
     public sealed record Location([property: Key] Guid Id, Guid OrganizationId, string Name);
 
-    public sealed record LocationUsers([property: Key] Guid LocationId, Guid[] UserIds);
+    public sealed record LocationUserAccess(
+        [property: Key] Guid UserId,
+        [property: Key] Guid LocationId
+    );
 
     public sealed record Family(
         [property: Key] Guid Id,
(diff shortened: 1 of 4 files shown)


src/CareTogether.Api/OData/LiveODataModelController.cs
```diff
@@ -25,6 +26,8 @@ namespace CareTogether.Api.OData
 {
     public sealed record Location([property: Key] Guid Id, Guid OrganizationId, string Name);
 
+    public sealed record LocationUsers([property: Key] Guid LocationId, Guid[] UserIds);
```
LarsKemmann, line 29: Actually better to make this flat, with a _compound key_:
```suggestion
    public sealed record LocationUserAccess([property: Key] Guid UserId, [property: Key] Guid LocationId);
```

LarsKemmann: The "Access" suffix helps suggest that in the future we might add `Role` or related info.

LarsKemmann (changes requested):

LarsKemmann (dismissed):

LarsKemmann (approved):

src/CareTogether.Api/OData/LiveODataModelController.cs
```diff
@@ -467,6 +484,28 @@ bool anonymize
...
+                            )
+                        )
+                    )
+                    .WhenAll()
+            )
+                .Where(person =>
+                    person.roles?.Contains(SystemConstants.ORGANIZATION_ADMINISTRATOR) ?? false
+                )
+                .Select(person => new LocationUserAccess(
+                    person.person.Id,
```
LarsKemmann, line 505: Instead of person ID, obtain the user ID for the person (from `familiesWithInfo`).

LarsKemmann (changes requested):

LarsKemmann (approved):

PR #883, opened 2025-05-06
Draft implementation of global api key

### LINKED ISSUE #875: Add a 'global' option for API key authentication & OData
- [ ] The ASP.NET Core authentication pipeline needs to recognize certain API keys as having global access to all data instead of to a specific organization.
- [ ] The ODataController needs to recognize a 'global' user and return data from *all* organizations, and from all locations in those organizations, if it is a global user.

PabloDinella:
Closes #875


PabloDinella:
Resulting data:
image

PR #884, opened 2025-05-07
Global odata api access

### LINKED ISSUE #875: Add a 'global' option for API key authentication & OData

PabloDinella:
Closes #875 

```shell
curl -X GET 'http://localhost:5000/api/odata/live/Family' \
  --user 'ffffffff-ffff-ffff-ffff-ffffffffffff:3f121e2f-16e8-43b7-8d6c-e58b11e91362'
```

### DIFF
--- a/src/CareTogether.Api/OData/LiveODataModelController.cs
+++ b/src/CareTogether.Api/OData/LiveODataModelController.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
+using System.Configuration;
 using System.Linq;
 using System.Threading.Tasks;
 using CareTogether.Engines;
@@ -24,16 +25,22 @@ using Timelines;
 
 namespace CareTogether.Api.OData
 {
+    public sealed record Organization([property: Key] Guid Id);
+
     public sealed record Location([property: Key] Guid Id, Guid OrganizationId, string Name);
 
     public sealed record LocationUserAccess(
         [property: Key] Guid UserId,
+        [property: Key] Guid OrganizationId,
         [property: Key] Guid LocationId
     );
 
     public sealed record Family(
         [property: Key] Guid Id,
-        [property: ForeignKey("LocationId")] Location Location,
+        // [property: ForeignKey("OrganizationId")] Organization Organization,
+        [property: ForeignKey("LocationId")]
+            Location Location,
+        Guid OrganizationId,
         Guid LocationId,
         string Name,
         string? PrimaryEmail,
@@ -211,7 +218,7 @@ namespace CareTogether.Api.OData
     [Authorize(Policies.ForbidAnonymous, AuthenticationSchemes = "Basic")]
     public class LiveODataModelController : ODataController
     {
-        private sealed record DataAccessScope(Guid OrganizationId, bool Anonymize);
+        private sealed record DataAccessScope(Guid OrganizationId, bool Anonymize, bool IsGlobal);
 
         private readonly IPoliciesResource policiesResource;
         private readonly IAccountsResource accountsResource;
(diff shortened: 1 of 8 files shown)


src/CareTogether.Api/appsettings.Development.json
```diff
@@ -42,5 +42,9 @@
   "Research": {
     "ApiKey": "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa",
     "OrganizationId": "11111111-1111-1111-1111-111111111111"
+  },
+  "GlobalApiAccess": {
+    "ApiKey": "3f121e2f-16e8-43b7-8d6c-e58b11e91362",
```
LarsKemmann, line 47: Can you use an all-letters API key so it's clear that this is a test value? Something like `eeeeeeee-...`

src/CareTogether.Api/Startup.cs
```diff
@@ -320,6 +320,25 @@ out var researchOrganizationId
                                     return;
                                 }
 
+                                if (
```
LarsKemmann, line 323: This whole basic authentication implementation is vulnerable to timing attacks. Can you add a random delay (`await Task.Delay(...)`) from between 50-100ms at the beginning of the `OnValidateCredentials` lambda (line 287)? See also https://learn.microsoft.com/en-us/dotnet/api/System.Random?view=net-9.0

src/CareTogether.Api/OData/LiveODataModelController.cs
```diff
@@ -24,16 +25,22 @@
...
     public sealed record LocationUserAccess(
         [property: Key] Guid UserId,
+        [property: Key] Guid OrganizationId,
         [property: Key] Guid LocationId
     );
 
     public sealed record Family(
         [property: Key] Guid Id,
-        [property: ForeignKey("LocationId")] Location Location,
+        // [property: ForeignKey("OrganizationId")] Organization Organization,
```
LarsKemmann, line 40: Why is this commented out?

src/CareTogether.Api/OData/LiveODataModelController.cs
```diff
@@ -362,17 +369,91 @@ private async Task<LiveModel> RenderLiveModelAsync()
...
-                    cacheEntry.SlidingExpiration = TimeSpan.FromMinutes(1);
-                    cacheEntry.AbsoluteExpiration = DateTimeOffset.Now.AddMinutes(5);
+            var organizationIds = isGlobal
+                ? await accountsResource.GetValidOrganizationsAsync()
+                : [organizationId];
+
+            var results = await organizationIds
+                .Select(async organizationId =>
+                    await cache.GetOrAddAsync(
+                        $"LiveODataModelController-RenderLiveModelAsync-{organizationId}-{Guid.NewGuid()}-{(anonymize ? "ANON" : "PII")}",
```
LarsKemmann, line 381: Why include the `{Guid.NewGuid()}` here? Won't that make it so the cache is *never* hit?

src/CareTogether.Api/OData/LiveODataModelController.cs
```diff
@@ -362,17 +369,91 @@ private async Task<LiveModel> RenderLiveModelAsync()
...
+                    Enumerable.Empty<Arrangement>(),
+                    Enumerable.Empty<ArrangementType>(),
+                    Enumerable.Empty<ChildLocationRecord>(),
+                    Enumerable.Empty<FamilyFunctionAssignment>(),
+                    Enumerable.Empty<IndividualFunctionAssignment>(),
+                    Enumerable.Empty<CommunityMemberFamily>(),
+                    Enumerable.Empty<CommunityRoleAssignment>()
+                ),
+                (acc, model) =>
+                    acc with
```
LarsKemmann, line 419: It's safer to write this code as `new LiveModel(...)` instead of `acc with { ... }` since the constructor syntax will force future developers to include all properties, e.g. if a new entity set is added to `LiveModel`.

src/CareTogether.Api/OData/LiveODataModelController.cs
PabloDinella: My editor pointed some syntax error and commented out to test another thing before fixing syntax, still uncomment in next commits, thanks for pointing out, this way I will remember

src/CareTogether.Core/Resources/Accounts/AccountsResource.cs
```diff
@@ -45,6 +45,21 @@ IEventLog<PersonAccessEvent> personAccessEventLog
             randomNumberGenerator = RandomNumberGenerator.Create();
         }
 
+        public async Task<Guid[]> GetValidOrganizationsAsync()
+        {
+            //WARNING: The read/write logic in this service needs to be designed carefully to avoid deadlocks.
```
PabloDinella, line 50: Is this warning relevant for this method? I kept it because I copied this method from `TryGetUserAccountAsync`, but not sure about the warning

LarsKemmann: It's just generally there for awareness with those reader/writer locks; when we rewrite this I'd like to clean all that up.

LarsKemmann (approved):

PR #890, opened 2025-05-13
Sort requested date

Rabi94:
Arrangement cards are now displayed from newest to oldest based on the requestedAtUtc date. If requestedAtUtc is missing, the fallback is startedAtUtc. Previously, the cards were sorted by startedAtUtc first, with requestedAtUtc as a secondary fallback.


LarsKemmann (dismissed):
@Rabi94 -- nice work, well done. I'm sorry it took me so long to get to this.

LarsKemmann (approved):

PR #900, opened 2025-05-21
Add Key attribute to OrganizationId and LocationId to all tables

PabloDinella:


src/CareTogether.Api/OData/LiveODataModelController.cs
LarsKemmann, line 1: `Location` also needs the composite key (there may be situations -- for data quality reasons -- where a location ID is not globally unique).

PabloDinella: Oops, overlooked that one

LarsKemmann (approved):

PR #902, opened 2025-05-28
Add ViewPersonDateOfBirth permission and return AgeInYears instead of…

PabloDinella:
… ExactAge when permission is not present


LarsKemmann (approved):
Nice solution -- I'm delighted that `AgeInYears` finally has a purpose! 😄

PR #910, opened 2025-06-04
Add AddEditOwnDraftNotes and DiscardOwnDraftNotes permissions

### LINKED ISSUE #908: Create a "Add/edit own draft notes" and "Discard own draft notes"
Currently, one can add/edit and discard all notes.

Related to support system ticket 11.

PabloDinella:
Fixes #908

### DIFF
--- a/src/CareTogether.Core/Permission.cs
+++ b/src/CareTogether.Core/Permission.cs
@@ -42,9 +42,10 @@
 
         AddEditDraftNotes = 180,
         DiscardDraftNotes = 181,
-        ApproveNotes = 182,
-        ViewAllNotes = 183, // Viewing your own notes is implicitly allowed
-
+        AddEditOwnDraftNotes = 182,
+        DiscardOwnDraftNotes = 183,
+        ApproveNotes = 184,
+        ViewAllNotes = 185, // Viewing your own notes is implicitly allowed
         // ---- Approvals ---- //
 
         ViewApprovalStatus = 200,
--- a/src/CareTogether.Core/Permission.cs
+++ b/src/CareTogether.Core/Permission.cs
@@ -42,10 +42,11 @@
 
         AddEditDraftNotes = 180,
         DiscardDraftNotes = 181,
-        AddEditOwnDraftNotes = 182,
-        DiscardOwnDraftNotes = 183,
-        ApproveNotes = 184,
-        ViewAllNotes = 185, // Viewing your own notes is implicitly allowed
+        ApproveNotes = 182,
+        ViewAllNotes = 183, // Viewing your own notes is implicitly allowed
+        AddEditOwnDraftNotes = 184,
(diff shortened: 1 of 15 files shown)


src/CareTogether.Core/Permission.cs
```diff
@@ -42,9 +42,10 @@ public enum Permission
 
         AddEditDraftNotes = 180,
         DiscardDraftNotes = 181,
-        ApproveNotes = 182,
```
LarsKemmann, line 45: We need to keep the integer values consistent for existing permissions, since _config.json_ files will otherwise give incorrect permissions for existing custom roles.

src/caretogether-pwa/src/Notes/NoteCard.tsx
```diff
@@ -32,6 +34,23 @@ export function NoteCard({ familyId, note }: NoteCardProps) {
 
   const permissions = useFamilyIdPermissions(familyId);
 
+  const userId = useLoadable(accountInfoState)?.userId;
+
+  const isOwnNote = note?.authorId === userId;
+
+  const canEditOwnNote =
```
LarsKemmann, line 41: Just a comment for future consideration as we design for V2: the backend should tell the client which permissions the client has for each object (basically _HATEOAS_ from REST -- or whatever the equivalent pattern is when we implement GraphQL?) and the UI can then just render controls for whichever actions are allowed by the server. Otherwise we have to keep implementing this business logic on the client and server, which is basically why the current client code is so messy and difficult to work with.

test/CareTogether.Core.Test/AuthorizationEngineTests/AuthorizeNoteCommandAsync.cs
```diff
@@ -0,0 +1,288 @@
...
+
+        [TestMethod]
+        public async Task UserHasNoAccessToNotes()
+        {
+            var user = PersonUserWithRoles(
+                Id('4'),
+                Guid.Parse("e3aaef77-0e97-47a6-b788-a67c237c781e")
+            );
+
+            mockUserAccessCalculation!
```
LarsKemmann, line 125: Can you create a helper method for this pattern? Something like:
```c#
private MockUserAccess(params IEnumerable<Permission> permissions)
{
    mockUserAccessCalculation!
        .Setup(x => ...)
        .ReturnsAsync(ImmutableList.Create(permissions));
}
```

LarsKemmann (changes requested):

test/CareTogether.Core.Test/AuthorizationEngineTests/AuthorizeNoteCommandAsync.cs
PabloDinella: Done

LarsKemmann (approved):

PR #911, opened 2025-06-06
Use autocomplete for adding permissions to a context in a Role  @PabloDinella

PabloDinella:
This will make UX for adding permissions easier.

User can select multiple permissions to add and click **ADD**.

Feature flag: `permissionsAutocomplete`

image


src/caretogether-pwa/src/Settings/Roles/PermissionsSelect.tsx
```diff
@@ -0,0 +1,62 @@
...
+          />
+        )}
+      />
+
+      <Button
+        variant="contained"
+        disabled={!value.length}
+        onClick={() => {
+          onAdd(value.map((v) => v.value));
+          setValue([]);
```
LarsKemmann, line 55: Double-check if this is needed.

PR #920, opened 2025-06-17
Fix infinite loading

### LINKED ISSUE #913: Infinite loading
Scenario:

1. Open a referral on a volunteer account
2. Remove permissions to view that referral
3. Refresh the referral page
4. "Loading family..." keeps showing indefinitely

A better experience would be to show a message saying the user can't access that referral/family. Current behavior also doesn't allow the user to click any menu item or logo, forcing him to manually change the URL to go to home, for example.

Rabi94:
I've fixed #913 the issue of the infinite loading when a user  has no permission to view a family, Now, instead of an infinite loading, the user will see:
"You don't have permission to view this family."
A Home button has also been added to allow the user to return to the Dashboard easily.
Screenshot (211)

This improves the user experience by providing a way to navigate without needing to manually change the URL.

please let me know if anything else should be adjusted!

### DIFF
--- a/src/caretogether-pwa/src/GeneratedClient.ts
+++ b/src/caretogether-pwa/src/GeneratedClient.ts
@@ -993,6 +993,10 @@ export class MetadataClient {
         this.baseUrl = baseUrl ?? "";
     }
 
+    /**
+     * Generates the OData $metadata document.
+     * @return The IEdmModel representing $metadata.
+     */
     getMetadata(): Promise<IEdmModel> {
         let url_ = this.baseUrl + "/api/odata/live/$metadata";
         url_ = url_.replace(/[?&]$/, "");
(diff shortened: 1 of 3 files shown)


Rabi94:
@PabloDinella I adjusted the font size to normal text and changed the message to: ‘Oops! You can’t view this family. It may be restricted or unavailable.’ & I’m displaying the message at the top of the page. 

Here’s a screenshot 

Screenshot (213)

PabloDinella (approved):

src/caretogether-pwa/src/GeneratedClient.ts
```diff
@@ -2016,10 +2016,10 @@ export enum Permission {
...
     AddEditDraftNotes = 180,
     DiscardDraftNotes = 181,
-    AddEditOwnDraftNotes = 182,
-    DiscardOwnDraftNotes = 183,
-    ApproveNotes = 184,
-    ViewAllNotes = 185,
+    ApproveNotes = 182,
+    ViewAllNotes = 183,
+    AddEditOwnDraftNotes = 184,
+    DiscardOwnDraftNotes = 185,
```
PabloDinella, line 2022: @LarsKemmann apparently I forgot to build the app after I reordered those items, so the generated files in branch `main` are behind the actual code :cold_sweat:

PabloDinella (approved):

LarsKemmann (approved):

PR #924, opened 2025-06-18
Unmasked Elements on Settings, Roles etc

Rabi94:
I already handled these in this PR:

* Settings page

* Role edit page

* Role creation sidepanel

* Dates in the arrangement card (requested at, started at, etc)

* Error page


LarsKemmann (dismissed):
We should apply the `ph-unmask` class just once, at the root page level, for anything in Settings or error screens.

PabloDinella (approved):

PR #934, opened 2025-07-02
Action definitions page

Rabi94:
@LarsKemmann 
Screenshot (1065)

Screenshot (1066)


LarsKemmann:
@Rabi94 With the new policy self-service UI work, let's plan to abandon this PR, but before we do that, please upload a screenshot or two of how the implementation looked last summer. It would be really good to have that comparison.

PabloDinella:
Screenshots for the newer self-service settings:

image
image

PR #935, opened 2025-07-04
Disabling OData during NSwag Swagger/OpenApi codegen to remove noise

LarsKemmann:
OData .NET assemblies on Windows seem to have XMLdoc comments included that Linux assemblies don't. I'm not 100% sure that that's the root cause, but anytime someone builds on Linux the descriptions are removed from OData-related Swagger/OpenApi schema elements. Those aren't needed for Swagger anyways since the SPA client doesn't ever call the OData endpoints, so this PR effectively turns off OData during NSwag codegen (which should also slightly help with the NSwag post-build step performance).

@PabloDinella, please try building locally to see if your machine changes the swagger.json file from the committed version or keeps it unmodified. If there are any other modifications then please check those in and we can troubleshoot to see where the remaining Windows-Linux differences are.


LarsKemmann:
This is the wrong approach. Trying with an NSwag document processor instead.

PR #1007, opened 2025-10-08
Merge Help Options into Single Help Bubble

Rabi94:
Merged “Feature requests” and “I have a problem” into one help bubble.


LarsKemmann:
Nice work Pablo. We can certainly live with that margin or padding flicker until the V2 UI shell. 😊

PR #1028, opened 2025-11-14
 Add complete other

### LINKED ISSUE #1018: Add "Complete other..." option

Rabi94:
The ‘Complete Other’ option was added to the volunteer card. When the 3 dots are clicked, a popup opens with an autocomplete, a Cancel button, and a Complete/Exempt button, a requirement (action) can be selected from the autocomplete and then completed or exempted.

Screenshot (500)

Screenshot (501)

Screenshot (502)

Screenshot (503)


PabloDinella (changes requested):
I think most of the new logic in FamilyScreen.tsx could go in a separate file, and be called from the AdultCard instead of FamilyScreen, could you make this change? 

Also, I think you don't need a state for person id, you can reference the person id directly from the props instead of saving it on a state. Probably the same for CompleteOtherContext.

PabloDinella (approved):

PR #1032, opened 2025-11-28
Performance improvements on authorization checks

PabloDinella:
While working with big datasets (4,000+ records), the /Records endpoint gets incredibly slow to respond.

This PR introduces some changes in the way we handle authorization and data disclosure, moving some lookups up in the call tree, so that those lookups are done only once per request, instead of being done on every family disclose and authorization check. 

This reduced the average time for the endpoint to respond on my machine, from ~4min to less than 10s:

## Before:

2025-11-27_22-43_before

## After:

2025-11-27_22-28_after

## Profilling before:
(lots of family lookups for user in session, for example)

image

## Profilling after:
(family lookup is made only a few times at the beginning)

image

### DIFF
--- a/src/CareTogether.Api/Startup.cs
+++ b/src/CareTogether.Api/Startup.cs
@@ -274,6 +274,7 @@ namespace CareTogether.Api
             );
             services.AddSingleton<IRecordsManager>(
                 new RecordsManager(
+                    policiesResource,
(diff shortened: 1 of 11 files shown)


LarsKemmann:
Just want to note here for posterity how thrilled I am by this, @PabloDinella...🔥

PR #1037, opened 2025-12-23
 V1Referrals

Rabi94:
This PR introduces V1 Referrals as a first-class, standalone domain, implemented end-to-end across the backend and frontend. The work establishes referrals as independent from cases while remaining fully aligned with the existing event-sourced architecture and frontend data patterns.

**Backend**

- A new event-sourced V1 Referrals aggregate was introduced, including referral state, events, commands, and resource wiring.
- Referrals are persisted and rehydrated exclusively from their event streams, ensuring auditability and consistency with existing aggregates.
- Referrals were intentionally kept separate from CombinedFamilyInfo. While a referral may reference a family, it is not owned by a family in the same way cases are. This reduces coupling and supports global intake, pre-assignment.
- A dedicated API endpoint was added to return all referrals (assigned and unassigned), allowing filtering to remain a frontend concern.
- Full DI and startup wiring was completed so referral commands flow through the same atomic records pipeline as other aggregates.
- The GeneratedClient was regenerated, producing typed referral DTOs, enums, and client methods required by the frontend.

**Frontend**

- A proper Recoil-based referrals model was implemented, including a query for loading referrals and command callbacks for write operations.
- ReferralDetailsPage to display the referral info
- V1Referrals page a table to display a table of the referrals ReferralRow, ReferralsFilter etc
- The Add New Referral drawer was fully wired to backend commands, supports referrals with unknown families, validates all required fields, and submits persisted data end-to-end.


PabloDinella (dismissed):
Good job!

The only changes I would make here for now are:

1. Don't create a new controller for referrals. Referrals are part of the Records concept. Records is everything related to the data that an organization owns.
2. Use react-hook-form for managing the form state and validation (maybe install zod for validation), that will help keeping the code more concise and consistent

src/CareTogether.Core/Engines/Authorization/AuthorizationEngine.cs
```diff
@@ -214,6 +215,102 @@ ArrangementsCommand command
...
+            locationId,
+            userContext,
+            new GlobalAuthorizationContext()
+        ),
+
+        UpdateV1ReferralFamily c => await userAccessCalculation.AuthorizeUserAccessAsync(
+            organizationId,
+            locationId,
+            userContext,
+            new FamilyAuthorizationContext(c.FamilyId)
```
LarsKemmann: This means that the user only needs to have access to the _updated_ family ID. But what if they don't have access to the _existing_ family ID? This would be a security failure.

package-lock.json
LarsKemmann: Delete this file

package.json
LarsKemmann: Delete this file

LarsKemmann (changes requested):

LarsKemmann (dismissed):

src/CareTogether.Core/Resources/V1Referrals/V1ReferralCloseReason.cs
```diff
@@ -0,0 +1,11 @@
+namespace CareTogether.Resources.V1Referrals
+{
+    public enum V1ReferralCloseReason
```
LarsKemmann: Make this a dynamic list from the organization configuration instead.

src/CareTogether.Core/Managers/Records/RecordsManager.cs
```diff
@@ -241,6 +267,49 @@ AtomicRecordsCommand command
 
             await ExecuteCommandAsync(organizationId, locationId, user, command);
 
+if (command is V1ReferralRecordsCommand referralCommand &&
+    referralCommand.Command is UpdateV1ReferralFamily update)
+{
+    var referrals = await v1ReferralsResource.ListReferralsAsync(
```
LarsKemmann: Replace with `GetReferralAsync(...)` instead

src/CareTogether.Core/Managers/Records/RecordsManager.cs
```diff
@@ -768,5 +868,57 @@ private Guid[] GetFamilyIdsFromCommand(AtomicRecordsCommand command) =>
                     $"The command type '{command.GetType().FullName}' has not been implemented."
                 ),
             };
+            private async Task<bool> FamilyHasOpenCaseAsync(
+    Guid organizationId,
+    Guid locationId,
+    Guid familyId
+)
+{
+    var cases = await v1CasesResource.ListV1CasessAsync(
```
LarsKemmann:
```suggestion
    var cases = await v1CasesResource.ListV1CasesAsync(
``` -- maybe?

File: src/CareTogether.Core/Resources/V1Referrals/V1Referral.cs [hunk 5]
```diff
@@ -0,0 +1,120 @@
...
+                    },
+
+                    V1ReferralAccepted accepted => referral with
+                    {
+                        Status = V1ReferralStatus.Accepted,
+                        FamilyId = accepted.FamilyId,
+                        AcceptedAtUtc = accepted.AcceptedAtUtc
+                    },
+
+                    V1ReferralClosed closed => ApplyClosed(referral, closed),
```
LarsKemmann: Don't make this a helper method since nothing else is using the helper method approach.

src/CareTogether.Core/Resources/V1Referrals/V1ReferralsResource.cs
```diff
@@ -0,0 +1,326 @@
...
+
+            await eventLog.AppendEventAsync(
+                organizationId,
+                locationId,
+                newEvent,
+                lastSequence
+            );
+        }
+
+        public async Task<V1Referral?> GetOpenReferralForFamilyAsync(
```
LarsKemmann: This method can be deleted, I think.

src/CareTogether.Core/Resources/V1Referrals/V1ReferralsResource.cs
```diff
@@ -0,0 +1,326 @@
...
+                organizationId,
+                locationId,
+                command.ReferralId
+            );
+
+            var referral = events.Any()
+                ? V1Referral.Rehydrate(events)
+                : null;
+
+            var newEvent = command switch
```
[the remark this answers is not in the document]
LarsKemmann: I think our call got cut off at this point. 😊 I would say address the other comments from the review, and then you can address this comment last. Feel free to talk through it with me. This might be a chance for us to do some pair programming.

PabloDinella (changes requested):
Don't create a controller for referral notes, referral notes should be loaded into the referral, like family notes are loaded into the family, all under the RecordsController.

PabloDinella:
Here are the new changes we talked on our last call:

- [x] Remove Incomplete/Completed from Family Screen only for client families (still needed for volunteer families)
- [x] Timeline entries merged, with an indicator of the referral the entry is coming from
- [x] Documents merged, with an indicator of the referral the document is coming from
- [x] Indicate incompleted, completed, exempted in the referral badge (show only non-zero counts)
- [x] Fix bug of "Close" referral not showing sometimes
- [x] Add "Link to existing case" for the scenario of linking to already-closed cases (for historical data entry)
- [x] Make the 'Automatically link & accept referrals' feature opt-in via a dialog
      Scenarios:
        1. When a Case is opened from the referral page, automatically connect this referral with the opened case
        2. When a Family with existing cases is selected in the referral page, automatically show a dialog asking if the user wants to link the referral with one of the existing cases
        3. When a new Family is created from the referral page, nothing else happens.

PabloDinella (approved):

PR #1051, opened 2026-01-28
support chat bubble

Rabi94:
Added a global bottom safe area in the root layout so content is never covered by the support chat widget, including Save buttons and other page actions. Extra padding is applied on mobile to account for the bottom navigation.


PabloDinella:
Made some adjustments: added `px` suffix (it was adding a huge space, as in the first image), adjusted RoleEdit page, and moved the widget a bit up on mobile, so it's not on top of the bottom navigation bar

image

image

PR #1068, opened 2026-04-08
Created Family and Individual Requirements table in Odata

marcusgraca:
Added FamilyRequirementStatus and IndividualRequirementsStatus in Odata. Both tables contains the following columns:

- OrganizationId
- LocationId
- FamilyId/ PersonId
- RequirementName
- Status
- Date
- ExpireDate

For exemptions, only the ExpireDate is populated.

### DIFF
--- a/src/CareTogether.Api/OData/LiveODataModelController.cs
+++ b/src/CareTogether.Api/OData/LiveODataModelController.cs
@@ -302,7 +302,8 @@ namespace CareTogether.Api.OData
         [property: Key] Guid PersonId,
         [property: Key] string RequirementName,
         string Status,
-        DateTimeOffset? Date
+        DateTimeOffset? Date,
+        DateTimeOffset? ExpireDate
     );
 
     public sealed record LiveModel(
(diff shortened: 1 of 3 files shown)


src/CareTogether.Api/OData/LiveODataModelController.cs
```diff
@@ -285,6 +286,25 @@ public sealed record RoleApproval(
...
     );
 
+    public sealed record FamilyRequirementStatus (
+        Guid OrganizationId,
+        Guid LocationId,
+        [property: Key] Guid FamilyId,
+        [property: Key] string RequirementName,
+        string Status,
+        DateTimeOffset? Date,
+        DateTimeOffset? ExpireDate
```
LarsKemmann, line 296: Why does this have an expire date and the individual one doesn't?

LarsKemmann (dismissed):
This looks good to me. I just have one question.

src/CareTogether.Api/OData/LiveODataModelController.cs
marcusgraca: It was my mistake. I updated the PR in a new commit.

LarsKemmann (approved):

PabloDinella (approved):

PR #1103, opened 2026-05-04
Fix Intake filter 

### LINKED ISSUE #1101: New client not appearing in INTAKE list on client page
This issue is linked to our feedback platform. For more details and updates, please visit [this link](https://caretogether.featurebase.app/p/new-client-not-appearing-in-intake-list-on-client-page).

Rabi94:
Fix Intake filter to include families with an open referral and no open case.

### DIFF
--- a/src/caretogether-pwa/src/V1Cases/PartneringFamilies.tsx
+++ b/src/caretogether-pwa/src/V1Cases/PartneringFamilies.tsx
@@ -21,7 +21,11 @@ import {
   UnfoldLess as UnfoldLessIcon,
   UnfoldMore as UnfoldMoreIcon,
 } from '@mui/icons-material';
-import { ArrangementPhase, Permission } from '../GeneratedClient';
+import {
+  ArrangementPhase,
+  Permission,
+  V1ReferralStatus,
+} from '../GeneratedClient';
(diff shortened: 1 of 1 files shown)


src/caretogether-pwa/src/V1Cases/PartneringFamilies.tsx
```diff
@@ -149,15 +167,28 @@ function PartneringFamilies() {
...
         switch (arrangementsFilter) {
           case 'All':
             return true;
 
           case 'Intake':
-            return !!openCase && arrangements.length === 0;
+            return (
+              (hasOpenReferralWithoutCase || !!openCase) &&
+              !arrangements.some((arrangement) =>
+                isSetupOrActiveArrangementPhase(arrangement.phase)
```
PabloDinella: This changes the existing Intake semantics for open cases. Previously, an open case only matched Intake when arrangements.length === 0, which aligns with the tooltip’s “no arrangements yet” wording. With this condition, a case whose arrangements are all Ended or Cancelled now matches because none are setup/active. I think we should preserve the prior arrangements.length === 0 behavior for 'Intake' case.

PabloDinella (changes requested):

PabloDinella (approved):

Issue #71, opened 2021-08-22
Fix startup performance


LarsKemmann:
- Ensure full auto-warmup of all configured organizations occurs on startup (before requests)
- Ensure Always On is configured in the App Service

LarsKemmann:
Always On has been enabled for the App Service, so this is now solely about eager loading of the data & eager event replay.

PR #1104, opened 2026-05-04
Handle signed-in users without organization access

PabloDinella:
## Summary
- Add a dedicated no-organization-access screen for signed-in users who are not connected to any CareTogether org/location
- Route post-login root traffic through a new RootRoute that redirects to a valid location or shows the no-access screen
- Guard scoped /org/:organizationId/:locationId routes so stale URLs from another account do not set invalid location context
- Extract access route helpers and screen components into src/caretogether-pwa/src/Access

localhost_3000_org_11111111-1111-1111-1111-111111111111_22222222-2222-2222-2222-222222222222_referrals

## Validation
- npm run type-check
- npx eslint src/AppRoutes.tsx src/Access/accessRouteHelpers.ts src/Access/NoOrganizationAccessScreen.tsx src/Access/RootRoute.tsx --ext ts,tsx --report-unused-disable-directives --max-warnings 0
- npm run build

### DIFF
--- a/src/caretogether-pwa/src/Access/NoOrganizationAccessScreen.tsx
+++ b/src/caretogether-pwa/src/Access/NoOrganizationAccessScreen.tsx
@@ -1,0 +1,63 @@
+import {
+  Box,
+  Button,
+  Container,
+  Paper,
+  Stack,
+  Typography,
+} from '@mui/material';
+import {
+  Logout as LogoutIcon,
+  Refresh as RefreshIcon,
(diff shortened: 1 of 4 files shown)


PabloDinella:
This was something a client pointed out (conversation #71 in featurebase), that when a new staff member signup in CareTogether, he sees this screen: 
image
This PR improves that experience, so the new user knows he has to wait for an invitation link to be able to access an organization/location.

PR #1106, opened 2026-05-11
Update Backend Packages

Rabi94:
1- Update .NET TO .NET10
2- Update NuGet packages for .NET 10 compatibility
3- Fix package migration issues across Azure Blob Storage, MSTest, and PowerBI APIs
4- Regenerate Swagger and frontend client from the updated backend build

***Just a heads up* there’s still a transitive vulnerability coming from Microsoft.Kiota.Abstractions through Microsoft.Graph
I tried forcing the newer Kiota version directly, but it’s not compatible with the current Graph SDK we’re using. It ended up breaking app startup and NSwag generation, so I rolled that back and kept the compatible dependency chain for now.**

I also tested the app on my end and it looks like it’s working fine. I tested referrals, cases, arrangements, adding adults, adding children, and adding families to communities.


src/CareTogether.Core/Utilities/FileStore/BlobFileStore.cs
```diff
@@ -42,7 +42,7 @@ string documentSubpath
                 $"{locationId}/{fileCategory}/{documentSubpath}"
             );
             var sasUri = objectBlob.GenerateSasUri(
-                BlobSasPermissions.Create,
+                BlobSasPermissions.Create | BlobSasPermissions.Write,
```
PabloDinella: Is this change needed?

Rabi94: I added this change while troubleshooting the document upload bug on my local machine and forgot to remove it afterward. It’s unrelated to the package updates.

PabloDinella: Ok!

PabloDinella (dismissed):

LarsKemmann (approved):

LarsKemmann:
@Rabi94 @PabloDinella Fun fact: merging this PR just reduced `/Records` load times for my production test customer from ~2.9s to 2.3s. :)

PabloDinella:
That's awesome!

PR #1113, opened 2026-05-18
 Add family sort options

Rabi94:
Add family sort options so users can view families by the default family-name order or by open date.
 
Screenshot (1023)

Screenshot (1024)


PabloDinella:
Did some updates:

- Fixed the alignment of the filters on the top (was broken since some PRs ago)
- Moved the sorting select one row below
- Added ascending and descending variations
- Fixed a cell span issue that was preventing arrangement cards fill the entire space
- Moved custom fields filters to a side panel just like in the Volunteers screen, to reduce visual pollution

image

PabloDinella (approved):

PR #1115, opened 2026-05-19
Added Role Name to the Requirements tables

marcusgraca:
To enable proper usability of the Requirements tab in the reports, we need to associate each requirement with its corresponding Role Name.

In the current model, the **Role Name** is only linked to requirements that are in a "Pending" state, returning NULL when the requirement has the status "Completed" or "Exempted".

This change adjusts the logic so that the **Role Name is also captured when a requirement is marked as Completed or Exempted**.

Additionally, the model in the OData Controller was updated to expose these columns in the FamilyRequirementStatuses and IndividualRequirementStatuses tables.


src/CareTogether.Core/Resources/Approvals/IApprovalsResource.cs
```diff
@@ -53,7 +53,8 @@ public sealed record CompleteVolunteerFamilyRequirement(
         string RequirementName,
         DateTime CompletedAtUtc,
         Guid? UploadedDocumentId,
-        Guid? NoteId
```
LarsKemmann, line 58: @PabloDinella is the issue here with this approach that a requirement can apply to more than one role? If so, then does this need to be redesigned, or is it correct as is?

PabloDinella:
Updated the implementation based on the review and follow-up design discussion.

The original PR direction added role information to the complete/exempt command path so future completed or exempted requirements would carry a role value. We changed that design because the command/event data is the durable source of truth, and role attribution is derived from policy/version context rather than something the user explicitly selects when marking a requirement complete or exempt.

The final shape is:

- Completion/exemption commands remain unchanged, so we do not change how requirement updates are stored going forward.
- `CompletedRequirementInfo` and `ExemptedRequirementInfo` now expose optional calculated `RoleNames` arrays on the read model.
- `PolicyEvaluationEngine.CalculateVolunteerFamilyApprovalsAsync` populates those role-name arrays for completed and exempted family/individual requirements from the existing role/policy calculation data.
- `CombinedFamilyInfoFormatter` passes the enriched calculated requirement lists through to `VolunteerFamilyInfo` / `VolunteerInfo`.
- OData no longer performs the completed/exempted role attribution itself; it only explodes the calculated `RoleNames` array into multiple rows.
- Pending statuses still use the existing calculated `MissingRequirements.Versions` role/version data.

This preserves all applicable roles for shared requirements without persisting an arbitrary first role on future command events.

PabloDinella (dismissed):

PabloDinella (approved):

PR #1123, opened 2026-05-21
 Fix referral closing timeline

### LINKED ISSUE #1118: Referral closing timeline entries always show midnight
Image

Rabi94:
Set today’s referral close time to the current time while keeping past close dates at 12:00 AM.
Screenshot (1032)
Use MUI DateTimePicker for referral close dates so today defaults to the current time and past dates default to 12:00 AM unless the user changes the time.
Screenshot (1035)

### DIFF
--- a/src/caretogether-pwa/src/Activities/ActivityTimeline.tsx
+++ b/src/caretogether-pwa/src/Activities/ActivityTimeline.tsx
@@ -94,6 +94,20 @@ const composeNoteType = (activity: Activity): string | null => {
   return null;
 };
 
+const shouldShowTimelineTime = (item: MergedTimelineItem) => {
+  if (item.kind !== 'family-activity') return false;
+
+  return item.activity instanceof ChildLocationChanged;
+};
+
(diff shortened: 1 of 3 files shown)


LarsKemmann:
Oh, I see the issue here. Thank you for doing this work, but I now realize that this is the wrong approach. We don't actually want users to have to put in the time of day here. I think that the correct fix would be to show only the date on the timeline, for all entries where we don't actually get a specific time. The only time that we want a time component is with a child location change. Every other requirement, note, etc. should not have that in the input field or in the timeline display.

I had been thinking about the timestamp on the timeline from the perspective of an audit log showing the exact time when a user put something into the system. I think that's confusing the issue, using the same date/time field on the timeline to mean two different things, so that was my mistake. At some point, we can create an audit log view or something like that, but for now, we should really keep that timeline simple and date-focused. And an audit log wouldn't allow customizing the timestamp anyways.

LarsKemmann (approved):

LarsKemmann:
@Rabi94 Can you please apply this fix to the V2 UI as well?

PR #1133, opened 2026-05-27
Upgrade MUI 7 to 9

Rabi94:
Upgraded the MUI to v9. Added required MUI v9 peer packages, updated MUI X Date Pickers, and fixed compatibility issues from removed/deprecated APIs.
Key changes include replacing removed GridLegacy imports with a local compatibility bridge, moving removed system props into sx, updating Autocomplete, Select, TextField, Checkbox, and Date Picker APIs, and keeping Timeline/Masonry isolated through app wrappers.


LarsKemmann:
Abandoning this since we should merge only this plus the following fixes.

PR #1135, opened 2026-05-28
Stabilization performance post-MUI 9

Rabi94:
This PR cleans up a few UI issues that showed up after the MUI v9 upgrade and also improves route loading performance across the app. I fixed some sidebar alignment problems caused by the new MUI spacing behavior and adjusted the referral timeline sizing so it feels more consistent with the family timeline layout, while keeping the changes small and targeted instead of rewriting large UI sections.
I also added **React.lazy** route splitting for larger screens like _Dashboard, Families, Referrals, Volunteers, Reports, Settings, and Support._ This helps reduce the initial bundle size, improves first-load responsiveness, and allows the main shell to render faster while individual pages load in the background.


LarsKemmann:
@Rabi94 Please fix the merge conflicts and then ping me so I can review & approve.

src/caretogether-pwa/playwright/.auth/admin.json
LarsKemmann, line 1: @Rabi94 Can/should this be removed & .gitignore'd?

src/caretogether-pwa/playwright-report/index.html
LarsKemmann, line 1: @Rabi94 This should be removed & future reports .gitignore'd.

src/caretogether-pwa/test-results/.last-run.json
LarsKemmann, line 1: Remove this and .gitignore the 'test-results/' directory

src/caretogether-pwa/src/AppRoutes.tsx
LarsKemmann, line 1: I don't like the full-page flash that we see now when loading the lazy modules. I do agree that lazy loading is nice to have, but it's not worth the disruption to the user when navigating between pages.

LarsKemmann (changes requested):

LarsKemmann (approved):

PR #1154, opened 2026-06-01
 Make family relationship field optional with free form autocomplete

Rabi94:
The field has been made optional when:

- Creating a new volunteer family
- Creating a new client family from referral flow
- Adding a new adult from the family screen
- Editing an adult’s family relationship

### DIFF
--- a/src/caretogether-pwa/src/Families/AddAdultDialog.tsx
+++ b/src/caretogether-pwa/src/Families/AddAdultDialog.tsx
@@ -7,6 +7,7 @@ import {
   DialogContent,
   DialogTitle,
   Divider,
+  Autocomplete,
   FormControl,
   FormControlLabel,
   FormGroup,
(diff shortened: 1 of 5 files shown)


LarsKemmann:
I'm going to hold off on merging this because I had a conversation with a customer that suggests we may want to prevent arbitrary data being entered, and just come up with a better set of default relationships for users to pick from (e.g., 'Single Mom', 'Mom + Partner', 'Mom + Wife')

LarsKemmann (changes requested):
I'm going to hold off on merging this because I had a conversation with a customer that suggests we may want to prevent arbitrary data being entered, and just come up with a better set of default relationships for users to pick from (e.g., 'Single Mom', 'Mom + Partner', 'Mom + Wife'). I'll circle back to this once I've followed up with the customer.

PR #1156, opened 2026-06-01
Policy Self-Service

Rabi94:


src/CareTogether.Core/Resources/Policies/EffectiveLocationPolicyValidator.cs
```diff
@@ -0,0 +1,551 @@
...
+                {
+                    if (IsBlank(alternateName))
+                    {
+                        errors.Add(
+                            $"Action Definition '{actionName}' cannot contain a blank alternate name."
+                        );
+                        continue;
+                    }
+
+                    if (actionDefinitions.ContainsKey(alternateName.Trim()))
```
LarsKemmann: This validation rule does not make sense. Alternate names will _almost always_ match the name of another action definition -- that's the whole point, in fact.

LarsKemmann (dismissed):

src/CareTogether.Core/Resources/Policies/EffectiveLocationPolicyValidator.cs
LarsKemmann: The same issue probably exists in the UI as well. :)

PabloDinella:
Updated the self-service policy gating in d1cf8a7d.

- Kept the existing Settings screens/routes visible.
- Gated only the new Location Settings policy/self-service tabs behind the PostHog flag.
- Renamed the flag key to `self-service-policy` via `SELF_SERVICE_POLICY_FEATURE_FLAG`.

Verified with `npm run type-check` and targeted ESLint for the touched files.

PabloDinella (approved):

PR #1164, opened 2026-06-08
Optimization records resource indexes no benchmarks

PabloDinella:
## Summary

Improves `/Records` performance by replacing repeated per-aggregate authorization and resource scans with request-scoped snapshots and model-maintained indexes.

The changes are based on the captured CPU profile and local benchmark runs against the same large records dataset.

## Changes

- Preload authorization inputs once per records request and evaluate family, referral, and community visibility from a `UserAccessCalculationSnapshot`.
- Reuse precomputed family and community permissions during disclosure instead of recalculating authorization.
- Create a family-rendering snapshot for notes, V1 cases, volunteer-family approvals, and referrals.
- Maintain secondary indexes for notes by family, V1 cases by family and assigned volunteer family, and referrals by family.
- Add targeted resource lookups so non-snapshot callers avoid full collection scans.
- Stream append-blob event logs asynchronously line by line instead of buffering each blob in memory.
- Document the profiling findings, implementation, expected impact, and remaining opportunities in `docs/records-authorization-performance.md`.

## Expected Impact

- Fewer full-tenant scans and immutable collection materializations during authorization.
- Fewer repeated approval, note, V1 case, and referral lookups while rendering families.
- One authorization pass for records visibility and disclosure where permissions can be reused.
- Lower peak memory use while replaying append-blob event logs.

## Benchmark Results

The benchmark exercised the same `/Records` endpoint and location for every run. Timings are from the `IRecordsManager.ListVisibleAggregatesAsync` boundary in `../tfg-optimization/benchmark-records-boundaries.csv`.

| Version | Runs | Mean | Range |
| --- | ---: | ---: | ---: |
| Baseline (`base-run-*`) | 4 | 63.390 s | 59.985-65.449 s |
| Authorization snapshot only (`auth*`) | 4 | 27.898 s | 27.125-29.513 s |
| Final indexed implementation (`indexes-*`) | 3 | 1.490 s | 1.467-1.525 s |

Compared with the baseline, the final implementation is **42.5x faster** and reduces measured records-list time by **97.65%**.

The resource indexes and rendering snapshot account for a further **18.7x improvement** over the authorization-snapshot-only version. In the final runs:

- `ListFamilyNotesAsync` calls fell from 5,757 to zero, replaced by one bulk `ListNotesAsync` call.
- `ListV1CasessAsync` calls fell from 22,281 to 2.
- `ListReferralsAsync` calls fell from 4,931 to 1.
- `ListLocationCommunitiesAsync` calls fell from 11,597 to 1.
- `AuthorizeUserAccessAsync` calls fell from 11,596 to zero, replaced by one request-scoped snapshot and synchronous evaluations.

## Verification

The branch was also verified with:

```bash
dotnet test test/CareTogether.Core.Test/CareTogether.Core.Test.csproj --no-restore -nologo /nodeReuse:false /maxcpucount:1
dotnet build CareTogetherCMS.sln --no-restore -nologo /nodeReuse:false /maxcpucount:1
```

The solution build regenerated the API/client artifacts through NSwag with no generated file changes.

### DIFF
--- a/docs/records-authorization-performance.md
+++ b/docs/records-authorization-performance.md
@@ -1,0 +1,83 @@
+# Records Authorization Performance
+
+## Background
+
+The `/Records` endpoint was profiled with
+`/home/pablo/github/CareTogether/tfg-optimization/app_perf.speedscope.json`.
+The profile showed that records listing spent most of its CPU time repeatedly
+recalculating authorization and rebuilding filtered immutable collections.
+
+The biggest hotspots were:
+
+- `UserAccessCalculation.AuthorizeUserAccessAsync`
+- repeated `ImmutableList.CreateRange` calls inside authorization
+- repeated `ApprovalsResource.TryGetVolunteerFamilyAsync` calls
+- repeated `CombinedFamilyInfoFormatter.RenderCombinedFamilyInfoAsync` calls
+- repeated family disclosure authorization from `AuthorizationEngine.DiscloseFamilyAsync`
+
+The main issue was not one expensive permission rule. It was the request shape:
+`RecordsManager.ListVisibleAggregatesAsync` authorized every family, referral,
+and community independently, and family rendering then authorized the same family
+again during disclosure.
+
+## What Changed
+
+### Request-Scoped Authorization Snapshot
+
+`UserAccessCalculation` now supports a `UserAccessCalculationSnapshot`.
+
+The snapshot preloads the tenant and user data needed by authorization once per
+records-list request:
+
+- current user family
+- V1 cases grouped by partnering family id
+- V1 cases grouped by assigned volunteer family id
+- referrals by referral id
+- volunteer family ids
+- communities
+- community ids by family id
+- current user's community role assignments
+- permission sets applicable to the user's configured roles
+
+`RecordsManager.ListVisibleAggregatesAsync` creates this snapshot once and then
+uses synchronous snapshot evaluation for family, referral, and community
+visibility checks.
+
+### Reused Family Permissions During Disclosure
+
+`AuthorizationEngine.DiscloseFamilyAsync` now has an overload that accepts
+precomputed family permissions.
+
+`CombinedFamilyInfoFormatter.RenderCombinedFamilyInfoAsync` accepts optional
+precomputed permissions and passes them into disclosure when available.
+
+`RecordsManager.ListVisibleAggregatesAsync` now carries the permissions from the
+family visibility check into family rendering. This avoids recalculating family
+authorization during disclosure for records-list rendering.
+
+## Expected Impact
+
+The change reduces repeated work in the `/Records` request path:
+
+- fewer full-tenant V1 case scans during authorization
+- fewer repeated approval lookups for the same family ids
+- fewer repeated community membership scans
+- fewer immutable list materializations inside authorization
+- one family authorization pass for records visibility and disclosure instead of two
+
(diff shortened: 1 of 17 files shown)


PabloDinella:
### Records List Flow

At a high level, the PR changes `/Records` from repeated per-aggregate resource
lookups into a bulk-load-and-reuse flow.

`RecordsManager.ListVisibleAggregatesAsync` now:

1. Creates the current `SessionUserContext`.
2. Loads families, the current location policy, communities, and referrals in
   parallel.
3. Builds one `UserAccessCalculationSnapshot` from the loaded records.
4. Builds one `CombinedFamilyRenderingSnapshot` for family rendering.
5. Uses synchronous snapshot authorization for each family, referral, and
   community.
6. Passes the already-computed family permissions into family disclosure.

The important behavior-preservation detail is that the old asynchronous
`AuthorizeUserAccessAsync` method still exists. It now builds a small snapshot
for the single authorization request and delegates to the same snapshot
evaluator used by the optimized `/Records` path.

### Authorization Snapshot

`UserAccessCalculationSnapshot` stores the data authorization repeatedly needed
before this change:

- the current user's family
- families by id
- V1 cases by owning family id
- V1 cases by assigned volunteer family id
- V1 referrals by referral id
- volunteer family ids
- communities
- community ids by member family id
- current user's community role assignments
- permission sets for the user's configured roles

`AuthorizeUserAccess` can then evaluate a `FamilyAuthorizationContext`,
`V1ReferralAuthorizationContext`, or `CommunityAuthorizationContext` from that
snapshot without reloading policies, communities, approvals, referrals, or V1
cases for each aggregate.

### Rendering Snapshot

`CombinedFamilyInfoFormatter` now has a `CombinedFamilyRenderingSnapshot`. It
pre-groups records needed while rendering visible family aggregates:

- notes by family id
- V1 cases by family id
- assigned V1 cases by volunteer family id
- volunteer family entries by family id
- linked referrals by family id

When the snapshot is supplied, family rendering pulls from those dictionaries.
When it is not supplied, the formatter falls back to resource methods such as
`ListFamilyNotesAsync`, `ListV1CasesForFamilyAsync`, and
`ListReferralsForFamilyAsync`. That keeps existing non-records callers working
while giving the records endpoint the faster path.

### Disclosure Reuse

`AuthorizationEngine.DiscloseFamilyAsync` and `DiscloseCommunityAsync` now have
overloads that accept precomputed permissions.

This matters because visibility and disclosure use the same contextual
permissions. In the records list, the family visibility pass already determines
the permissions for each visible family. Passing those permissions into
`RenderCombinedFamilyInfoAsync` lets disclosure reuse them instead of calling
authorization again for the same family.

### Resource Indexes

The resource models now maintain in-memory secondary indexes during event
replay and command commits:

- `NotesModel` indexes note ids by family id.
- `V1CaseModel` indexes V1 case ids by owning family id.
- `V1CaseModel` also indexes V1 case ids by assigned volunteer family id.
- `V1ReferralModel` indexes referral ids by family id.

These indexes do not change persisted events or storage format. They are rebuilt
from the existing event logs when the resource models initialize, then updated
whenever a command commits.

### Event Log Streaming

`AppendBlobEventLog` now reads append-blob event logs line by line with a
`StreamReader` instead of reading the entire blob into memory as one string and
then wrapping it in a `StringReader`.

That is a supporting memory-pressure improvement for model initialization,
especially when large tenant event logs are replayed.

LarsKemmann (approved):

PR #1172, opened 2026-06-17
Fixes ChildOver18 alerts

### LINKED ISSUE #1158: Approved/onboarded volunteer families' children don't get flagged as over 18 in the inbox
@PabloDinella, I'm assigning you since you worked on https://github.com/CareTogether/CareTogetherCMS/pull/1157.

Rabi94:
Fixed ChildOver18 alerts for Volunteer and Partnering Family children.
Removed the volunteer-family restriction and updated the age check to include children who are exactly 18 years old (>= 18).
Verified alerts appear for  Volunteer and Partnering Family children, and that non-qualifying children do not generate alerts.
Frontend-only change. No backend or permission updates.


src/caretogether-pwa/src/Model/QueueModel.ts
```diff
@@ -39,8 +39,7 @@ const childrenOver18Query = selector<ChildOver18[]>({
   get: ({ get }) => {
     const visibleFamilies = get(visibleFamiliesQuery);
     return visibleFamilies
-      ?.filter((family) => family.volunteerFamilyInfo)
```
LarsKemmann, line 42: I'm double-checking with the user who reported this to see if this should actually be removed. Also, I think we may need to only alert on families that currently have a volunteer family role that is *not* set to inactive.

LarsKemmann (changes requested):

LarsKemmann (approved):

PR #1196, opened 2026-06-25
Approvals in family screen V2

### LINKED ISSUE #1192: Approvals redesign in family screen

Rabi94:
Screenshot (1174)
Screenshot (1175)
Screenshot (1176)


LarsKemmann:
Notes from our call:

- [x] Move the complete, exempt, etc. buttons to the top of the requirements drawer to the right of the status or stage badge. 
- [x] I'm unable to do anything other than complete or exempt missing application requirements when in my test customer tenant (as an admin)
- [x] Don't show removed role cards for individuals who have been removed from a family role (still show the overall family role if that was removed, and still show individual roles if those were removed.)
- [x] Fix the logic of the people who are shown in the participants within a family role. In a family role, we should see all of the active participants (in addition to the removed ones that are showing up), and not see the single row called 'Family' that is marked active.
- [x] In individual roles, don't show participants in the drawer since they are not relevant there. You can show the participation status of the individual person if you want, but that would apply at the individual role level. There wouldn't be multiple participants, so there's no need for that section.

LarsKemmann (approved):

PR #1232, opened 2026-07-10
 Family screen V2: Add active arrangement summary cards to case header

Rabi94:
Screenshot (1274)


PabloDinella:
Added a commit to fix ph-unmask usage #1229

PabloDinella (approved):

PR #1240, opened 2026-07-17
 V2 Clients Page

Rabi94:
This PR introduces the new V2 Clients Browser page using the shared V2 architecture and MUI X DataGrid.

V2

Screenshot (1283)

V1
Screenshot (1281)


PabloDinella (changes requested):
Before merging we need to work on filters performance. Filtering by name with 800+ rows practically crashes my browser.

Rabi94:
@PabloDinella The performance issue wasn't caused by the name filter itself, but by the amount of work performed on every keystroke. Each search update triggered the entire ViewModel pipeline, including filtering, sorting, rebuilding presentation rows, computing assignment/custom field display values, and forcing the DataGrid to process a brand-new set of row objects.
To address this, I split the pipeline into two stages. Presentation rows are now built and cached independently from filtering, allowing search to operate on cached display data instead of rebuilding it for every keypress. I also introduced useDeferredValue to keep typing responsive while React schedules the more expensive filtering work, and stabilized several DataGrid props to reduce unnecessary re-renders.
The overall behavior, permissions, sorting, and filtering logic remain unchanged, but the amount of synchronous work performed during search has been significantly reduced, resulting in a much more responsive experience for larger organizations.

PabloDinella:
After testing it with a customer's data it's still very slow, I'm looking into what can be done to improve it further

PabloDinella (approved):

PR #1273, opened 2026-07-29
 Improve Arrangements table columns and default sorting

Rabi94:
This PR includes displaying the actual arrangement type, splitting the Duration column into separate planned/actual date columns, and updating the default workflow sorting.
Screenshot (1316)


LarsKemmann (approved):
This looks good! @Rabi94 I still want to fix this so that it shows all arrangements, not just the ones from a "selected" case (we still need to get rid of that concept), but that's a separate issue.

PR #1320, opened 2026-09-11
Upgrade Clients with MUI DataGrid Premium

Rabi94:
Upgrades the V2 Clients experience to use **MUI DataGrid Premium** as the primary
browsing, filtering, sorting, and reporting interface, The Clients grid now provides native filtering, column management, Pivot,
Charts, and export capabilities while dynamically adapting to each
organization's configured fields and assignment roles.

Users can now:

1. Search and filter Clients
2. Show or hide relevant organization fields
3. Build a Pivot using eligible fields
4. Aggregate using Client count
5. Visualize the results with Charts
6. Export data through the native DataGrid tools


PabloDinella (approved):
Very good UX, looking forward for this toolbar on the other datagrids too :)

